PR #21, opened 2026-02-26
implement sessions

OpakAlex:


internal/orchestrator/orchestrator.go
```diff
@@ -41,49 +45,92 @@ type noopParser struct{}
...
+	sessions SessionStoreInterface,
 	customRules []string,
 	contentPreparers []ContentPreparerEntry,
 	luaScriptPaths map[string]string,
+	permissionChecker PermissionChecker,
+	permissionPluginName string,
+	summarizeAfterMessages int,
+	maxMessagesAfterSummary int,
+	summarizePrompt string,
+	summarizeUpdatePrompt string,
```
zhisme: `NewWithRules` now takes **14 positional parameters**. Every call site (including all tests) is a wall of `nil, "", 0, 0, "", ""`. This is hard to read and easy to get wrong.

Consider an options struct:
```go
type OrchestratorOpts struct {
    PermissionChecker       PermissionChecker
    PermissionPluginName    string
    SummarizeAfterMessages  int
    MaxMessagesAfterSummary int
    SummarizePrompt         string
    SummarizeUpdatePrompt   string
}
```

internal/state/store/session.go
```diff
@@ -0,0 +1,174 @@
...
+		Messages:  []provider.Message{},
+		Metadata:  map[string]string{},
+		CreatedAt: time.Now(),
+		UpdatedAt: time.Now(),
+	}
+}
+
+// AddMessage appends a message and persists. If maxMessages > 0, trims to last maxMessages.
+func (s *SessionStore) AddMessage(id string, msg provider.Message) error {
+	sess, err := s.Get(id)
```
zhisme: `AddMessage` does a full `Get()` (SELECT + JSON unmarshal of all messages), appends one message, re-marshals everything, and UPDATEs. Two problems:

1. **Performance**: O(n) JSON work per message. For sessions with 50+ messages this adds up.
2. **Race condition**: No transaction wraps the read-modify-write. Two concurrent requests to the same session can both read N messages, each append one, and the second UPDATE overwrites the first — losing a message.

At minimum, wrap this in a transaction with `BEGIN IMMEDIATE` so the second writer blocks until the first commits.

internal/state/store/memory.go
```diff
@@ -0,0 +1,139 @@
...
+// MemoriesForContext returns memories visible to the current actor: all general (actor_id IS NULL)
+// plus all for actor.Actor(ctx), optionally filtered by tag. Empty tag means no filter.
+func (s *MemoryStore) MemoriesForContext(ctx context.Context, tag string) ([]*state.Memory, error) {
+	actorID := actor.Actor(ctx)
+	query := `SELECT id, actor_id, content, tags, created_at FROM memories WHERE (actor_id IS NULL OR actor_id = ?)`
+	args := []interface{}{actorID}
+	if tag != "" {
+		query += ` AND (tags LIKE ? OR tags LIKE ? OR tags LIKE ?)`
+		// JSON array: "tag", "tag" at start, or in middle
+		args = append(args, `%"`+tag+`"%`, `"`+tag+`"%`, `%,"`+tag+`"%`)
```
zhisme: Tag filtering via `LIKE` with `%"tag"%` is fragile — it matches substrings. For example, searching for tag `"work"` will match `["workflow"]` because the string contains `"work"`.

Consider using `json_each()` (supported by modernc/sqlite):
```sql
AND EXISTS (SELECT 1 FROM json_each(tags) WHERE json_each.value = ?)
```
Or store tags in a separate join table.

internal/orchestrator/orchestrator.go
```diff
@@ -215,13 +262,14 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string) (
 	}); err != nil {
 		return nil, fmt.Errorf("adding user message: %w", err)
 	}
+	o.maybeSummarizeSession(ctx, sessionID)
```
zhisme: `maybeSummarizeSession` is called synchronously **before** the agent loop. This makes an LLM call that blocks the user's request. If summarization takes 2-3 seconds, that's added latency the user feels on top of the normal LLM response time.

Consider running this asynchronously (goroutine) or deferring it to after the response is sent.

internal/state/store/db.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	if err := os.MkdirAll(dataDir, 0700); err != nil {
+		return nil, fmt.Errorf("state store: %w", err)
+	}
+	dbPath := filepath.Join(dataDir, "state.db")
+	db, err := sql.Open("sqlite", dbPath+"?_journal_mode=WAL")
+	if err != nil {
+		return nil, fmt.Errorf("state store: open db: %w", err)
+	}
+	if _, err := db.Exec("PRAGMA journal_mode=WAL"); err != nil {
```
zhisme, line 38: WAL mode is enabled but there is no `PRAGMA busy_timeout`. With concurrent connections (multiple channels writing to the DB), SQLite will return `SQLITE_BUSY` immediately instead of retrying. Standard practice:
```go
db.Exec("PRAGMA busy_timeout = 5000")
```

internal/state/store/session.go
```diff
@@ -0,0 +1,174 @@
...
+
+// NewSessionStore returns a session store that uses the given DB.
+// maxMessages caps messages per session (0 = no cap); maxIdleDays enables pruning of idle sessions (0 = off).
+func NewSessionStore(db *DB, maxMessages, maxIdleDays int) *SessionStore {
+	return &SessionStore{db: db, maxMessages: maxMessages, maxIdleDays: maxIdleDays}
+}
+
+// Get loads a session by id. Returns error if not found.
+func (s *SessionStore) Get(id string) (*state.Session, error) {
+	var messagesJSON, summary, activeModel, metadataJSON, createdAt, updatedAt string
```
zhisme, line 28: The `summary` column in the migration is `TEXT` (nullable), but here it's scanned into a plain `string`. If the DB has a SQL NULL for summary (e.g. from a manual insert or future migration), this may fail depending on the driver.

Use `sql.NullString` or `COALESCE(summary, '')` in the query.

internal/orchestrator/orchestrator.go
```diff
@@ -418,13 +470,24 @@ func (o *Orchestrator) executeCall(ctx context.Context, call ToolCall) ToolResul
 		}
 	}
 
+	actorID := actor.Actor(ctx)
+	if actorID != "" && o.permissionChecker != nil && call.Plugin != o.permissionPluginName {
+		allowed, err := o.permissionChecker.Allowed(ctx, actorID, call.Plugin)
+		if err != nil || !allowed {
```
zhisme: When the permission check returns an error, it is silently treated as denial with a generic `"permission denied"`. The actual error is swallowed — no logging. This will be painful to debug in production.

At minimum:
```go
if err != nil {
    log.Printf("Warning: permission check for actor %s plugin %s: %v", actorID, call.Plugin, err)
}
```

internal/state/store/memory.go
```diff
@@ -0,0 +1,139 @@
...
+		if tagsJSON != "" {
+			_ = json.Unmarshal([]byte(tagsJSON), &tags)
+		}
+		t, _ := time.Parse(time.RFC3339, createdAt)
+		out = append(out, &state.Memory{ID: id, Content: content, Tags: tags, CreatedAt: t})
+	}
+	return out
+}
+
+func mustParseTime(s string) time.Time {
```
zhisme: Nit: `mustParseTime` is named `must*` (Go convention = panics on error) but silently returns zero time on parse failure. Rename to `parseTimeOrZero` or similar to avoid confusion.

cmd/opentalon/main.go
```diff
@@ -70,9 +71,31 @@ func main() {
...
-	sessions := state.NewSessionStore(dataDir)
-	memory := state.NewMemoryStore(dataDir)
-	_ = memory.Load()
+	var memory orchestrator.MemoryStoreInterface
+	var sessions orchestrator.SessionStoreInterface
+	if dataDir != "" {
+		db, err := store.Open(dataDir)
+		if err != nil {
+			log.Printf("Warning: state store open: %v; using in-memory state", err)
+			mem := state.NewMemoryStore("")
```
zhisme: The in-memory fallback block (lines 80-83) is duplicated in the `else` branch below (lines 94-97). Extract a helper:
```go
func newInMemoryState() (orchestrator.MemoryStoreInterface, orchestrator.SessionStoreInterface) {
    mem := state.NewMemoryStore("")
    _ = mem.Load()
    return mem, state.NewSessionStore("")
}
```

internal/state/store/plugin_migrations.go
```diff
@@ -0,0 +1,105 @@
...
+			return err
+		}
+		if err := tx.Commit(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func pluginMigrationNumber(name string) (int, error) {
```
zhisme: `pluginMigrationNumber` is identical to `migrationNumber` in `db.go`. Extract into a shared helper to avoid duplication.

zhisme (changes requested):
## Review Summary

This PR bundles three significant features: SQLite-backed persistent state, actor-scoped memories with a permission plugin, and session summarization. The SQLite store and migration framework are solid, and the test coverage for the store layer is good. However, there are architectural concerns, correctness issues, and missing test coverage that should be addressed before merge.

### Key issues

1. **`NewWithRules` has 14 positional parameters** — unreadable and fragile. Needs an options struct.
2. **`AddMessage` does unprotected read-modify-write** — race condition under concurrent requests to the same session; no transaction wrapping.
3. **Tag filtering via LIKE is fragile** — substring matches cause false positives (e.g. tag `"work"` matches `"workflow"`).
4. **Summarization blocks the user request** — synchronous LLM call in the hot path adds latency.
5. **No `PRAGMA busy_timeout`** — concurrent DB access will get `SQLITE_BUSY` errors.
6. **`summary` column nullable but scanned into plain `string`** — potential scan failure.
7. **No tests for permission checker, summarization, or actor context package.**
8. **PR bundles 3 concerns** — SQLite store, actor/permissions, and summarization ideally would be separate PRs for easier review and bisect.

### Minor

- `pluginMigrationNumber` and `migrationNumber` are identical — deduplicate.
- `Search` / `SearchByTag` on SQLite `MemoryStore` appear unused by the orchestrator now (only `MemoriesForContext` is called) — dead code?
- `summary` column in migration has no default, but `Create` inserts empty string `""` — inconsistent with nullable column.
- Duplicated in-memory fallback code in `main.go`.

PR #23, opened 2026-03-05
support install skills by prompt

OpakAlex:


pkg/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+			return parts[0] + "/" + parts[1], parts[1]
+		}
+		return "", ""
+	}
+	// org/repo
+	if strings.Contains(url, "/") {
+		parts := strings.SplitN(url, "/", 2)
+		if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+			name = parts[1]
+			return parts[0] + "/" + parts[1], name
```
zhisme: **Security (high): Path traversal in `parseInstallURL`**

The `org/repo` fallback doesn't validate that `name` is a clean path component. Input like `../../etc/evil` produces `name = "../../etc/evil"`, which is passed to `bundle.EnsureSkillDir` as a directory name.

`filepath.Join(stateDir, "skills", name)` resolves `..` components, so this can escape the skills directory. Even if the git clone fails (bad URL), `os.MkdirAll` may create directories outside `data_dir` before that.

All three branches (HTTPS, SSH, bare `org/repo`) have this issue — `parts[1]` can contain `/` or `..` since `SplitN(url, "/", 2)` only splits on the first `/`.

**Fix**: Validate that both `org` and `repo` contain no `/`, `..`, or path separators. Or reject if `filepath.Base(name) != name`:
```go
if strings.ContainsAny(name, "/\\.") || filepath.Base(name) != name {
    return "", ""
}
```

cmd/opentalon/main.go
```diff
@@ -200,7 +224,23 @@ func main() {
...
+	}
+
+	contentPreparers := make([]orchestrator.ContentPreparerEntry, 0, 1+len(cfg.Orchestrator.ContentPreparers))
+	// Commands preparer runs first so slash commands are handled before the LLM
+	contentPreparers = append(contentPreparers, orchestrator.ContentPreparerEntry{
+		Plugin:   "opentalon-commands",
+		Action:   "prepare",
+		ArgKey:   "text",
+		Insecure: false, // trusted: can return invoke for opentalon actions
+	})
```
zhisme: **Bug: Duplicate preparer execution**

`opentalon-commands` is hardcoded here as the first content preparer. But `config.example.yaml` also lists it under `content_preparers` (line 48). Anyone following the example config will run the slash command parser **twice** per message.

For non-slash messages both runs pass through (wasteful but harmless). For slash commands the first returns early via `runInvokeSteps`, so the second never executes — but the duplication is confusing and error-prone.

**Fix**: Either remove it from `config.example.yaml`'s `content_preparers` (since it's hardcoded here), or don't hardcode it and rely solely on config. I'd lean toward config-only — hardcoding couples the core to a specific external plugin name.

config.example.yaml
```diff
@@ -44,6 +44,10 @@ orchestrator:
   # Run these plugin actions before the first LLM call; their output becomes the user message (or they can block with send_to_llm: false).
   # List order = execution order: the first entry runs first and receives the user message; each preparer's output is the next one's input.
   content_preparers:
+    # Slash commands (/install skill, /show config, /commands, /set prompt, /clear) — run first
+    - plugin: opentalon-commands
```
zhisme, line 48: This duplicates the hardcoded `opentalon-commands` preparer in `main.go:238`. With this config, the slash command parser runs twice per message. See comment on `main.go`.

If you keep the hardcoded entry, remove this from the example. If you keep this, remove the hardcoded one.

pkg/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+	if e.cfg == nil {
+		return orchestrator.ToolResult{Content: "(config not available)"}
+	}
+	// Redact secrets for display
+	redacted := redactConfig(e.cfg)
+	data, err := yaml.Marshal(redacted)
+	if err != nil {
+		return orchestrator.ToolResult{Error: fmt.Sprintf("marshal config: %v", err)}
+	}
+	return orchestrator.ToolResult{Content: string(data)}
```
zhisme: **Bug: Missing `CallID` in returned `ToolResult`**

`showConfig` and `listCommands` both discard the `call` parameter (`_ orchestrator.ToolCall`) and return `ToolResult` without setting `CallID`. Every other action in this file (`installSkill`, `setPrompt`, `clearSession`) correctly sets `CallID: call.ID`.

If anything downstream matches results to calls by ID (e.g. for logging, display, or chaining), these results will be orphaned.

**Fix**: Accept `call orchestrator.ToolCall` instead of `_` and set `CallID: call.ID` on all returned `ToolResult` values (including error paths at lines 163 and 169).

internal/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (e *Executor) clearSession(call orchestrator.ToolCall) orchestrator.ToolResult {
+	sessionID := call.Args["session_id"]
+	if sessionID == "" {
+		return orchestrator.ToolResult{CallID: call.ID, Error: "session_id not set (internal error)"}
+	}
+	if err := e.sessions.Delete(sessionID); err != nil {
+		return orchestrator.ToolResult{CallID: call.ID, Error: fmt.Sprintf("delete session: %v", err)}
+	}
+	e.sessions.Create(sessionID)
```
zhisme, line 223: **Bug: `clearSession` doesn't clear `pendingPipelines`**

The orchestrator maintains `pendingPipelines[sessionID]` for multi-step pipeline confirmations (`orchestrator.go:102`). When `/clear` runs, this code deletes and re-creates the session but the stale pipeline entry survives in the map. The next message will hit the pending pipeline check (`orchestrator.go:212`) and try to confirm/cancel the old (now meaningless) pipeline.

The executor doesn't have access to this map. Consider:
- Exposing a `ClearPendingPipeline(sessionID)` method on the orchestrator
- Or having the orchestrator handle session clearing internally instead of delegating to the executor

internal/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+func (e *Executor) setPrompt(call orchestrator.ToolCall) orchestrator.ToolResult {
+	text := strings.TrimSpace(call.Args["text"])
+	if e.runtimePromptPath == "" {
+		return orchestrator.ToolResult{CallID: call.ID, Error: "runtime prompt path not configured"}
+	}
+	dir := filepath.Dir(e.runtimePromptPath)
+	if err := os.MkdirAll(dir, 0750); err != nil {
+		return orchestrator.ToolResult{CallID: call.ID, Error: fmt.Sprintf("create dir: %v", err)}
+	}
+	if err := os.WriteFile(e.runtimePromptPath, []byte(text), 0600); err != nil {
```
zhisme, line 205: **Security (medium): `/set prompt` is a global prompt injection surface**

The runtime prompt file is:
- **Global** — not scoped to a session or actor
- **Writable** by any chat user via `/set prompt <text>`
- **Injected** into the LLM system prompt for **all** subsequent requests (see `orchestrator.go:402-408`)

In multi-tenant deployments (e.g. shared Slack workspace), any user can modify the LLM's behavior for everyone. Even in single-user mode, this is an unrestricted injection surface — the text goes straight into the system prompt without any sanitization.

Consider:
- Scoping the prompt per-session or per-actor
- Requiring admin/owner permissions
- At minimum, length-limiting the text

internal/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+		cfg:               cfg,
+		runtimePromptPath: runtimePromptPath,
+	}
+}
+
+// Execute implements orchestrator.PluginExecutor.
+func (e *Executor) Execute(call orchestrator.ToolCall) orchestrator.ToolResult {
+	switch call.Action {
+	case ActionInstallSkill:
+		return e.installSkill(context.Background(), call)
```
zhisme, line 74: **Design: Detached context for network I/O**

`context.Background()` means the git clone inside `installSkill` → `bundle.EnsureSkillDir` can't be cancelled and ignores request deadlines/timeouts. The `PluginExecutor` interface doesn't pass context (existing limitation), but `installSkill` is the first action that does significant network I/O.

Worth a TODO or consider extending `PluginExecutor.Execute` to accept `context.Context`.

internal/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/opentalon/opentalon/internal/bundle"
+	"github.com/opentalon/opentalon/internal/config"
+	"github.com/opentalon/opentalon/internal/orchestrator"
+	"github.com/opentalon/opentalon/internal/requestpkg"
```
zhisme, line 13: **Design: `pkg/` package imports `internal/` packages**

Conventionally `pkg/` signals a stable public API boundary. This package imports four `internal/` packages (`bundle`, `config`, `orchestrator`, `requestpkg`), tightly coupling it to internal implementation details.

This works in Go (same module), but is surprising. Should this live in `internal/commands/` instead? It's not something external consumers would import.

internal/orchestrator/orchestrator.go
```diff
@@ -481,6 +520,27 @@ func (o *Orchestrator) executeCall(ctx context.Context, call ToolCall) ToolResul
...
+				if provide := o.contextArgProviders[name]; provide != nil {
+					if v := provide(ctx, name); v != "" {
+						args[name] = v
+					}
+				}
+			}
+			call.Args = args
+			break
+		}
+	}
```
zhisme: **Nit: Context arg injection runs before permission check**

This block runs before the permission check at line ~545. Functionally harmless (it just reads from the Go context), but logically the permission check should come first — there's no point injecting args for a call that's about to be denied.

Consider swapping the order: permission check first, then inject context args.

pkg/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+	out := *c
+	if out.Models.Providers == nil {
+		return &out
+	}
+	provs := make(map[string]config.ProviderConfig)
+	for name, p := range out.Models.Providers {
+		p2 := p
+		if p2.APIKey != "" {
+			p2.APIKey = "[redacted]"
+		}
```
zhisme: **Security (low): Shallow redaction**

Only `APIKey` fields on model providers are redacted. Other potentially sensitive config values — plugin configs (which may contain webhook secrets, tokens), internal service URLs, custom prompts — are shown verbatim via `/show config`.

Consider also redacting:
- Plugin `config` maps
- Any field named `secret`, `token`, `password`, etc.
- Or switch to a whitelist approach: only show known-safe fields

internal/commands/executor.go
```diff
@@ -0,0 +1,235 @@
...
+	if ref == "" {
+		ref = "main"
+	}
+
+	github, name := parseInstallURL(url)
+	if github == "" || name == "" {
+		return orchestrator.ToolResult{CallID: call.ID, Error: "could not parse url: use https://github.com/org/repo or org/repo"}
+	}
+
+	skillDir, err := bundle.EnsureSkillDir(ctx, e.dataDir, name, github, ref)
```
zhisme, line 106: **Security (medium): Any chat user can install arbitrary GitHub repos**

Skills are request packages — they define HTTP request templates (URL, body, headers). A malicious or compromised skill repo could define actions that:
- Hit internal services (SSRF)
- Exfiltrate data via HTTP requests
- Register misleading action names that shadow existing actions

The skill is immediately registered in the tool registry AND persisted to disk — no approval step, no allowlist, no confirmation.

Consider:
- An allowlist of trusted GitHub orgs/repos
- Requiring admin/owner actor permissions for `/install skill`
- A confirmation step before installing
- At minimum, logging the installation with actor ID for audit

zhisme (changes requested):
## Review Summary

### Critical
- **Path traversal** in `parseInstallURL` — `name` can contain `../` sequences that escape the skills directory (see inline)
- **No tests** — 307 lines of new code across `pkg/commands/executor.go` (235 lines) and `internal/config/installed_skills.go` (72 lines) with zero test coverage. `parseInstallURL` especially needs tests for edge cases and path traversal attempts. The new `actor.WithSessionID`/`SessionID` context helpers and the `InjectContextArgs` orchestrator logic are also untested.

### Bugs
- **Duplicate preparer execution** — `opentalon-commands` is both hardcoded in `main.go` and listed in `config.example.yaml`, causing it to run twice per message
- **Missing `CallID`** in `showConfig` and `listCommands` results
- **`clearSession` doesn't clear `pendingPipelines`** — stale pipeline entries survive session clear

### Security
- **`/set prompt`** writes a global file injected into the system prompt for ALL requests — prompt injection surface, especially in multi-tenant
- **`/install skill`** allows any chat user to clone arbitrary GitHub repos and register them as tools (HTTP request templates)
- **`redactConfig`** only redacts API keys — other config values may contain sensitive data

### Design
- `pkg/commands/` imports `internal/` packages — should probably be `internal/commands/`
- `context.Background()` in `installSkill` — git clone can't be cancelled
- `InjectContextArgs` injection runs before the permission check (nit)

zhisme (approved):
Looks good — all critical and high-severity items from the previous review are resolved.

### Remaining notes (good-to-fix, not blocking)

These are acknowledged in the code with TODOs or are design decisions. Flagging for visibility:

- **`clearSession` doesn't clear `pendingPipelines`** — TODO at `executor.go:431`. If a user runs `/clear` while a pipeline confirmation is pending, the stale pipeline survives. Worth wiring up a `ClearSession` hook on the orchestrator eventually.
- **`context.Background()` in `installSkill`** — TODO at `executor.go:285`. Git clone can't be cancelled. Would be nice to extend `PluginExecutor.Execute` to accept context when the interface is next touched.
- **`/set prompt` is still global** — Size-limited (32KB) now, which helps. In multi-tenant setups any user can still modify the system prompt for everyone. Per-session or per-actor scoping would be a good follow-up.
- **`/install skill` has no permission gate** — Audit logging is a good first step. An allowlist of trusted orgs or admin-only restriction would harden this for team deployments.
- **Two `GetCapability` + action-loop lookups in `executeCall`** (`orchestrator.go:543` and `566`) — Could be collapsed into one, very minor.

PR #28, opened 2026-03-09
feat: add MCP bridge support for HTTP+SSE MCP servers

OpakAlex:
- Add MCPServerConfig to pkg/requestpkg/types.go (public, importable by the external opentalon-mcp plugin binary)
- Add MCP *MCPServerConfig field to requestpkg.Set; skill YAMLs with an mcp: section are skipped by the HTTP executor (tools come from the plugin binary)
- Add CollectMCPServers() to requestpkg loader to gather MCP server configs from all loaded skill sets
- Add SetEnv() to plugin.Process and Env []string to plugin.PluginEntry so the MCP plugin binary receives OPENTALON_MCP_SERVERS at launch
- Add ServeListener() to pkg/plugin/server.go for TCP (remote gRPC) mode
- Wire it all together in main.go: request sets are loaded before plugins so MCP configs can be injected into the mcp plugin entry's environment

### DIFF
--- a/cmd/opentalon/main.go
+++ b/cmd/opentalon/main.go
@@ -232,15 +232,7 @@ func main() {
 				if e.Name != "mcp" {
 					continue
 				}
-				env := os.Environ()
-				filtered := make([]string, 0, len(env)+1)
-				for _, v := range env {
-					if !strings.HasPrefix(v, "OPENTALON_MCP_SERVERS=") {
-						filtered = append(filtered, v)
-					}
-				}
-				filtered = append(filtered, "OPENTALON_MCP_SERVERS="+string(mcpJSON))
-				pluginEntries[i].Env = filtered
+				pluginEntries[i].WithEnvOverride("OPENTALON_MCP_SERVERS", string(mcpJSON))
 				injected = true
 				break
 			}
--- a/cmd/opentalon/main.go
+++ b/cmd/opentalon/main.go
@@ -229,18 +229,20 @@ func main() {
 		} else {
 			injected := false
 			for i, e := range pluginEntries {
-				if e.Name == "mcp" {
-					env := os.Environ()
+				if e.Name != "mcp" {
+					continue
+				}
+				env := os.Environ()
 				filtered := make([]string, 0, len(env)+1)
-				for _, e := range env {
-					if !strings.HasPrefix(e, "OPENTALON_MCP_SERVERS=") {
-						filtered = append(filtered, e)
+				for _, v := range env {
+					if !strings.HasPrefix(v, "OPENTALON_MCP_SERVERS=") {
+						filtered = append(filtered, v)
 					}
 				}
-				pluginEntries[i].Env = append(filtered, "OPENTALON_MCP_SERVERS="+string(mcpJSON))
-					injected = true
-					break
-				}
+				filtered = append(filtered, "OPENTALON_MCP_SERVERS="+string(mcpJSON))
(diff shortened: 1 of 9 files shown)


cmd/opentalon/main.go
```diff
@@ -229,6 +219,40 @@ func main() {
...
+
+	// Collect MCP server configs and inject into the "mcp" plugin entry's environment
+	// before the plugin is launched so the binary knows which servers to connect to.
+	if mcpServers := requestpkg.CollectMCPServers(requestSets); len(mcpServers) > 0 {
+		mcpJSON, err := json.Marshal(mcpServers)
+		if err != nil {
+			log.Printf("Warning: marshal MCP servers: %v", err)
+		} else {
+			injected := false
+			for i, e := range pluginEntries {
```
zhisme, line 231: **Potential duplicate env var**: if `OPENTALON_MCP_SERVERS` is already set in the parent environment, `append(os.Environ(), ...)` creates a duplicate key. Most systems use the last value, but it's platform-dependent.

Should filter it out first:
```go
env := os.Environ()
filtered := make([]string, 0, len(env)+1)
for _, e := range env {
    if !strings.HasPrefix(e, "OPENTALON_MCP_SERVERS=") {
        filtered = append(filtered, e)
    }
}
pluginEntries[i].Env = append(filtered, "OPENTALON_MCP_SERVERS="+string(mcpJSON))
```

pkg/requestpkg/types.go
```diff
@@ -0,0 +1,9 @@
+package requestpkg
+
+// MCPServerConfig holds the configuration for one MCP server, as specified
+// in the mcp: section of a skill YAML file.
+type MCPServerConfig struct {
+	Server  string            `json:"server" yaml:"server"`
+	URL     string            `json:"url" yaml:"url"`
```
zhisme, line 7: **No validation on `MCPServerConfig`**: `Server` and `URL` are plain strings with no validation at load time. A skill YAML with `mcp:` but empty `url:` silently produces a broken config that gets serialized to the plugin env.

At minimum, `LoadDir`/`LoadSkillDir` should reject sets where `MCP != nil` but `URL == ""`.

pkg/plugin/server.go
```diff
@@ -20,6 +20,15 @@ type Handler interface {
 	Execute(req Request) Response
 }
 
+// ServeListener starts a gRPC server on an existing listener. The caller is
+// responsible for printing the handshake line to stdout before calling this.
+// Useful for TCP mode (MCP_GRPC_PORT).
```
zhisme, line 26: `Serve()` defers `ln.Close()` + `os.RemoveAll()`, but `ServeListener` doesn't close the listener. Listener cleanup responsibility is unclear — either document explicitly that the caller must close it, or add `defer ln.Close()` for consistency with `Serve()`.

zhisme (changes requested):
## Summary

Code changes are clean and well-scoped. A few issues to address:

- **No tests**: 100 lines added across 7 files, 0 test files touched. `CollectMCPServers` and the `Register` skip-MCP logic are both easily unit-testable — should have coverage.
- The `Env []string` comment on `PluginEntry` says "replaces inherited env" — accurate, but could trip up future callers who set it to `[]string{"FOO=bar"}` expecting append behavior. Consider a helper like `WithEnvOverride(key, value string)` instead of raw slice assignment.
- Type alias comment on `MCPServerConfig` is redundant — Go devs understand type aliases.

PR #29, opened 2026-03-10
support channel webhooks

OpakAlex:


zhisme (commented):
_(Superseded by inline review below)_

zhisme (commented):
## Verdict

Core feature is solid and well-tested. **Main blocker:** the race condition with unsynchronized goroutines in the webhook handler (see inline on `yaml_webhook.go`). The double JSON parse and server state reset are worth addressing too.

### Design notes (not tied to specific lines)

- **Single-port limitation:** The "only one port" constraint is fine for MVP, but the error on port conflict is a runtime surprise. Consider validating at config load time so the user gets a clear error before anything starts.
- **Inline transport branching:** `Start()` branches via `if wh := ch.spec.Inbound.HTTPWebhook; wh != nil`. Works now, but will get messy if more transport types are added.
- **No test for `Shutdown`:** There's no test verifying that `Shutdown` actually stops the server or handles in-flight requests gracefully.
- **`yaml_spec.go` `Regex` field:** Backwards-compatible (defaults to `false`), but worth a note in changelog.

internal/channel/yaml_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+			log.Printf("yaml-channel: %s: webhook %s %s from %s body=%s",
+				ch.spec.ID, r.Method, r.URL.Path, r.RemoteAddr, body)
+		}
+
+		// Respond immediately (Teams requires fast ack)
+		w.WriteHeader(responseCode)
+
+		// Process in goroutine
+		go ch.processInboundData(body)
+	}
```
zhisme, line 75: **Bug (HIGH): Race condition** — `go ch.processInboundData(body)` is fired without `ch.wg.Add(1)` / `defer ch.wg.Done()`. When `Stop()` is called, `ch.wg.Wait()` won't wait for in-flight webhook requests. Messages could be dropped or panic writing to a closed `inbox` channel.

internal/channel/yaml_ws.go
```diff
@@ -172,6 +173,18 @@ func (ch *YAMLChannel) handleFrame(conn *websocket.Conn, data []byte) {
 		}
 	}
```
zhisme, line 174: **Bug (MEDIUM): Double JSON parse on WebSocket path** — The original `handleFrame` already does `json.Unmarshal(data, &frame)` before calling this. After the refactor, `processInboundData` parses the JSON *again*. Every WebSocket message is now unmarshaled twice. The ack logic uses the first parse, event processing uses the second — wasteful and could diverge.

internal/channel/webhook_server.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}()
+	}
+
+	return nil
+}
+
+// Shutdown gracefully stops the webhook server.
+func (s *WebhookServer) Shutdown(ctx context.Context) error {
+	s.mu.Lock()
```
zhisme, line 71: **Bug (LOW-MEDIUM): Server state never reset after Shutdown** — After `Shutdown()`, `started` remains `true` and `mux` still has old routes. If channels are hot-reloaded, new registrations will fail because `ListenAndServe` on the old server is done but `started` is still `true`. Reset `started`, `mux`, `server`, and `port` here.

internal/channel/webhook_jwt.go
```diff
@@ -0,0 +1,256 @@
...
+
+// fetchKeys retrieves the JWKS from the OIDC discovery endpoint.
+func (v *JWTValidator) fetchKeys(ctx context.Context) error {
+	// Fetch OIDC discovery doc
+	req, err := http.NewRequestWithContext(ctx, "GET", v.oidcURL, nil)
+	if err != nil {
+		return fmt.Errorf("build OIDC request: %w", err)
+	}
+	resp, err := v.httpClient.Do(req)
+	if err != nil {
```
zhisme, line 177: **Bug (LOW): No HTTP status check** — Both the OIDC discovery fetch and the JWKS fetch read the body regardless of HTTP status. A 404 or 500 will produce a confusing "parse OIDC config" error. Check `resp.StatusCode` before parsing.

internal/channel/webhook_jwt.go
```diff
@@ -0,0 +1,256 @@
...
+		return fmt.Errorf("fetch OIDC config: %w", err)
+	}
+	body, _ := io.ReadAll(resp.Body)
+	_ = resp.Body.Close()
```
zhisme, line 181: **Nit:** `body, _ := io.ReadAll(resp.Body)` silently ignores read errors. Same on the line below for JWKS fetch. At minimum log them.

internal/channel/yaml_ws.go
```diff
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
```
zhisme, line 14: **Security (MEDIUM): Potential regex DoS** — `pattern` comes from YAML config via `substituteTemplate`. If any part of the pattern is derived from user/event context, a malicious input could craft a catastrophic backtracking regex. If patterns are always author-controlled (admin YAML only), document that assumption. Otherwise consider compiling once at load time or adding safeguards.

internal/channel/yaml_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+	"net/http"
+	"os"
+)
+
+// startWebhookInbound registers the HTTP webhook handler and starts the
+// shared webhook server. Returns immediately (server runs in background).
+func (ch *YAMLChannel) startWebhookInbound(wh *WebhookInboundSpec) error {
+	// Create JWT validator if configured
+	if wh.ValidateJWT {
+		audience := substituteTemplate(wh.Audience, ch.buildContexts())
```
zhisme, line 15: **Security (LOW): No warning when JWT validation is disabled** — If `validate_jwt` is false or omitted, this webhook endpoint is completely open to the internet with zero auth. Should log a warning at startup like "webhook endpoint /api/messages has no authentication configured".

internal/channel/yaml_webhook_test.go
```diff
@@ -0,0 +1,401 @@
...
+		t.Error("expected error when registering on a different port, got nil")
+	}
+
+	if s.server != nil {
+		_ = s.server.Close()
+	}
+}
+
+func TestRegisterWebhookRoute_DefaultPort(t *testing.T) {
+	s := newTestWebhookServer()
```
zhisme, line 229: **Test issue: This test is a no-op** — It calls `s.register(0, ...)` on a local instance (bypassing `RegisterWebhookRoute`) and then checks `if port <= 0 { port = 3978 }` inline in the test itself. This tests the test's own logic, not the production code's default port behavior.

internal/channel/yaml_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+
+		// Read body with 1MB cap
+		body, err := io.ReadAll(io.LimitReader(r.Body, 1<<20))
+		if err != nil {
+			log.Printf("yaml-channel: %s: read webhook body: %v", ch.spec.ID, err)
+			http.Error(w, "bad request", http.StatusBadRequest)
+			return
+		}
+		_ = r.Body.Close()
+
```
zhisme, line 60: **Nit:** `_ = r.Body.Close()` after `io.ReadAll` is unnecessary — `http.Server` closes the request body automatically.

internal/channel/webhook_jwt_test.go
```diff
@@ -0,0 +1,325 @@
...
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	if err := v.ValidateRequest(req); err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+}
+
+// splitJWT splits a JWT into its three dot-separated parts.
+func splitJWT(token string) []string {
+	parts := make([]string, 3)
```
zhisme: **Nit:** This 14-line hand-rolled string splitter does exactly what `strings.SplitN(token, ".", 3)` does.

PR #30, opened 2026-03-10
guards

OpakAlex:


internal/orchestrator/orchestrator.go
```diff
@@ -380,6 +392,11 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string) (
 
 		messages := o.buildMessages(ctx, sess, content)
 
+		messages, err := o.runGuardPlugins(ctx, messages)
+		if err != nil {
+			return &RunResult{Response: err.Error()}, nil
+		}
+
 		if os.Getenv("LOG_LEVEL") == "debug" {
 			log.Printf("[LLM request] round %d, %d messages:", i+1, len(messages))
```
zhisme: **Security: fail-open is dangerous for a security guard.**

When `executeCall` returns an error, you log a warning and `continue` — letting the potentially malicious content through unsanitized. The whole point of this feature is to block prompt injection; if the guard is down, you're flying blind.

This should fail-closed by default (block the request when the guard errors). Consider adding a `fail_open: bool` config field for non-critical guards where skipping is acceptable:

```go
if toolResult.Error != "" {
    log.Printf("Warning: guard %s.%s error: %s", g.Plugin, g.Action, toolResult.Error)
    if !g.FailOpen {
        return nil, fmt.Errorf("Guard %s.%s failed: %s", g.Plugin, g.Action, toolResult.Error)
    }
    continue
}
```

Same issue applies to the Lua guard error path and the missing-plugin `continue` above.

internal/orchestrator/orchestrator.go
```diff
@@ -176,7 +187,8 @@ func NewWithRules(
```
zhisme, line 175: **Using `error` for control flow here is a code smell.**

A guard blocking a request is not an error — it's a valid, expected outcome. But this returns `fmt.Errorf`, and then `Run()` does `err.Error()` as the response. If a *real* error occurs (nil pointer, panic recovery, etc.), it would also be surfaced to the user as a friendly "blocked" message, hiding the actual bug.

Consider returning a struct instead:
```go
type guardResult struct {
    blocked  bool
    message  string
    messages []provider.Message
}
```

internal/orchestrator/orchestrator.go
```diff
@@ -168,6 +171,14 @@ func NewWithRules(
 	if pipelineCfg.MaxStepRetries == 0 && pipelineCfg.StepTimeout == 0 {
 		pipelineCfg = pipeline.DefaultConfig()
 	}
+	var preparers, guards []ContentPreparerEntry
+	for _, p := range opts.ContentPreparers {
+		if p.Guard {
+			guards = append(guards, p)
+		} else {
```
zhisme, line 179: The Lua branch and plugin-executor branch here are near-identical copies of the same logic in the content preparer loop in `Run()`. Both do: check `lua:` prefix → `RunPrepare` → check `SendToLLM`, or check `HasAction` → build `ToolCall` → `executeCall` → parse `preparerResponse`.

Extract a shared helper like `runSinglePreparer(ctx, entry, content) (newContent, blockMsg string, blocked bool, err error)`. Would reduce this ~80-line function to ~15 lines and eliminate the risk of the two copies drifting apart over time.

zhisme (commented):
## Review: Guards

Good feature — guard plugins running before every LLM call is the right architecture for prompt injection defense. Documentation and test coverage are strong. A few issues below, one of which is a security concern for a feature whose job is literally security.

internal/orchestrator/orchestrator.go
```diff
@@ -98,6 +100,7 @@ type Orchestrator struct {
```
zhisme, line 97: `messages, err :=` here shadows the outer `messages` from `buildMessages`. This compiles fine but makes it easy to accidentally use the pre-guard `messages` below if someone refactors. A different variable name (e.g. `guardedMsgs`) would be clearer about the transformation.

internal/orchestrator/orchestrator.go
```diff
@@ -450,6 +467,86 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string) (
 	return nil, fmt.Errorf("agent loop exceeded %d iterations", maxAgentLoopIterations)
 }
 
+// runGuardPlugins runs all guard plugins on the last non-system message before an LLM call.
+// Guards sanitize content to prevent prompt injection from tool results or user input.
+// Returns the (possibly modified) messages, or an error containing the block message if a guard
+// returned send_to_llm: false.
+func (o *Orchestrator) runGuardPlugins(ctx context.Context, messages []provider.Message) ([]provider.Message, error) {
+	if len(o.guards) == 0 {
+		return messages, nil
```
zhisme: This allocates a new slice via `copy` on every call, even when no guard modified the content. Minor, but since this runs before every LLM call in the agent loop, you could track a `modified` bool and skip the copy when nothing changed.

internal/orchestrator/orchestrator.go
```diff
@@ -22,11 +22,13 @@ const maxAgentLoopIterations = 20
 const PermissionAction = "check"
 
 // ContentPreparerEntry configures a plugin action to run before the first LLM call.
+// If Guard is true, the plugin also runs before every LLM call in the agent loop to sanitize messages and prevent prompt injection.
 type ContentPreparerEntry struct {
```
zhisme, line 26: The comment still says "before the first LLM call" but this struct now also serves guard plugins that run before *every* LLM call. The added line helps but the original comment is now misleading — consider updating it to cover both cases.

internal/orchestrator/orchestrator_test.go
```diff
@@ -540,6 +540,257 @@ func (e *errorReturningExecutor) Execute(call ToolCall) ToolResult {
...
+	exec := &prefixingExecutor{prefix: "SAFE:"}
+	llm := &capturingLLM{responses: []string{"all good"}}
+	parser := &fakeParser{parseFn: func(string) []ToolCall { return nil }}
+	guards := []ContentPreparerEntry{{Plugin: "guard-plugin", Action: "sanitize", Guard: true}}
+
+	orch := setupGuardOrchestrator(guards, llm, parser, map[string]PluginExecutor{"guard-plugin": exec})
+	result, err := orch.Run(context.Background(), "s1", "user message")
+	if err != nil {
+		t.Fatal(err)
+	}
```
zhisme, line 608: This test validates that guard errors are skipped (fail-open). If/when the fail-open behavior is changed to fail-closed by default (see comment on `runGuardPlugins`), this test would need to be updated to expect a blocked response instead.

PR #41, opened 2026-03-16
support chrome

OpakAlex:


cmd/opentalon/main.go
```diff
@@ -261,6 +261,20 @@ func main() {
 		}
 	}
 
+	// Inject chrome plugin config from the plugin's config: block in config.yaml.
+	for i, e := range pluginEntries {
+		if e.Name != "chrome" {
```
zhisme: **Unnecessary plugin-specific hardcoding**

Every `PluginEntry` already has a `Config map[string]interface{}` loaded from YAML. This PR takes that generic config, marshals it to JSON, and injects it as an env var — but only for `"chrome"`. There's nothing chrome-specific about this operation. Any plugin could need its config passed via env var.

The existing MCP injection at least has a justification: it collects MCP servers from *request sets* (a cross-cutting concern across the config). The chrome PR just round-trips the plugin's own `.Config` field — that's a generic operation dressed up as a special case.

**Should be generalized.** Instead of hardcoding `if e.Name != "chrome"`, this should be a generic loop for all plugins with a non-empty Config:

```go
for i, e := range pluginEntries {
    if len(e.Config) == 0 {
        continue
    }
    cfgJSON, err := json.Marshal(e.Config)
    if err != nil {
        log.Printf("Warning: marshal %s config: %v", e.Name, err)
        continue
    }
    pluginEntries[i].WithEnvOverride("OPENTALON_PLUGIN_CONFIG", string(cfgJSON))
}
```
That way main.go stays plugin-agnostic and any plugin binary can read its own config without changes to the core.

cmd/opentalon/main.go
```diff
@@ -261,6 +261,20 @@ func main() {
...
+			continue
+		}
+		chromeJSON, err := json.Marshal(e.Config)
+		if err != nil {
+			log.Printf("Warning: marshal chrome config: %v", err)
+			break
```
zhisme: **Minor code issues**

Uses `break` after the inner `break` on error — correct but the double-break-in-loop pattern is a smell. A `continue` on error + single `break` after the success path would be cleaner. Also no nil check on `e.Config` before marshaling — a plugin with no config block would get `OPENTALON_CHROME_CONFIG=null` injected into its environment.

cmd/opentalon/main.go
```diff
@@ -261,6 +261,20 @@ func main() {
...
+		}
+		pluginEntries[i].WithEnvOverride("OPENTALON_CHROME_CONFIG", string(chromeJSON))
```
zhisme: **The `Config` field is unused for plugins today**

`Config` is loaded from YAML for all plugin entries but currently only consumed by *channels* (via `ConfigurableChannel` interface). This PR implicitly establishes that plugins can also use `Config`, but does so through a backdoor (env var injection in main.go) rather than a proper interface.

zhisme (changes requested):
### Issues

See inline comments for details on each point:

**1. Unnecessary plugin-specific hardcoding** — the `Config` round-trip is a generic operation, not chrome-specific.

**2. Should be generalized** — a single loop for all plugins with non-empty Config.

**3. Minor code issues** — double-break pattern smell and missing nil check on `e.Config`.

**4. The `Config` field is unused for plugins today** — this implicitly establishes plugin config usage through a backdoor rather than a proper interface.

zhisme (commented):
test

pkg/plugin/grpc_server.go
```diff
@@ -13,7 +13,12 @@ type grpcServer struct {
 	handler Handler
 }
 
-func (s *grpcServer) Capabilities(_ context.Context, _ *emptypb.Empty) (*pluginpb.PluginCapabilities, error) {
+func (s *grpcServer) Capabilities(_ context.Context, req *pluginpb.PluginInitRequest) (*pluginpb.PluginCapabilities, error) {
+	if c, ok := s.handler.(Configurable); ok {
+		if err := c.Configure(req.GetConfigJson()); err != nil {
```
zhisme, line 20: **`Configure` errors are swallowed**

If `Configure()` returns an error it's logged but the Capabilities response is still returned normally. A plugin with broken config will silently proceed with no/bad configuration. Consider either returning the error (failing the handshake) or at minimum surfacing it to the host so it can decide whether to proceed.

pkg/plugin/grpc_server.go
```diff
@@ -13,7 +13,12 @@ type grpcServer struct {
 	handler Handler
 }
 
-func (s *grpcServer) Capabilities(_ context.Context, _ *emptypb.Empty) (*pluginpb.PluginCapabilities, error) {
```
zhisme, line 16: **Config delivery is coupled to the Capabilities RPC**

`Capabilities` now serves two purposes: "tell me what you can do" _and_ "here's your config." These are semantically different concerns. If `Capabilities` is ever called a second time (e.g. during `Reload`), the plugin gets re-configured as a side effect. A dedicated `Init` or `Configure` RPC would separate these concerns cleanly, but this is a minor design point — not a blocker if you want to keep it simple for now.

zhisme (changes requested):
### Issues addressed from previous review

The chrome-specific hardcoding is gone. Config is now passed generically through the gRPC `Capabilities` RPC via a new `PluginInitRequest` message, with an opt-in `Configurable` interface for plugin handlers. This is the right approach.

### Remaining issues

See inline comments for details:

**1. `Configure` errors are swallowed** — plugin with broken config will silently proceed.

**2. Config delivery is coupled to the Capabilities RPC** — mixes two semantically different concerns.

PR #42, opened 2026-03-17
add concurrency

OpakAlex:


internal/orchestrator/orchestrator.go
```diff
@@ -341,9 +359,45 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
+	select {
+	case o.semaphore <- struct{}{}:
+		defer func() { <-o.semaphore }()
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+
+	// Per-session lock: serializes concurrent messages for the same session.
+	sm := o.acquireSessionLock(sessionID)
+	defer o.releaseSessionLock(sessionID, sm)
```
zhisme, line 405: **Bug: Race condition on `pendingPipelines`**

`pendingPipelines` is a plain `map[string]*pipeline.Pipeline` that was previously protected by the global `o.mu`. This PR removes `o.mu` and replaces it with per-session locks, but Go maps are not safe for concurrent access even on **different keys**. Two goroutines running `Run()` for different sessions will concurrently read/write `pendingPipelines` (lines ~354 and ~403 in the old file), which can panic at runtime.

**Fix:** Protect `pendingPipelines` with a dedicated `sync.Mutex`, or use `sync.Map`.

internal/channel/registry.go
```diff
@@ -118,6 +125,10 @@ func (r *Registry) StopAll() {
 func (r *Registry) dispatch(ch pkg.Channel, inbox <-chan pkg.InboundMessage) {
 	defer r.wg.Done()
 
+	sem := make(chan struct{}, r.maxConcurrent)
```
zhisme: **Double concurrency control is redundant**

Both `Registry.dispatch()` (here) and `Orchestrator.Run()` now have their own semaphores with the same `maxConcurrent` value. Since the handler calls `Run()`, one of these is redundant — the effective limit is `min(N, N) = N`, but you pay for two semaphore acquisitions per request.

**Suggestion:** Pick one layer to own concurrency control. The orchestrator is the better place since it also owns per-session serialization. The registry can remain a simple fan-out without its own semaphore.

internal/channel/registry.go
```diff
@@ -126,15 +137,29 @@ func (r *Registry) dispatch(ch pkg.Channel, inbox <-chan pkg.InboundMessage) {
...
+
+			// Acquire a dispatch slot (blocks when at maxConcurrent).
+			select {
+			case sem <- struct{}{}:
+			case <-r.ctx.Done():
+				return
 			}
+
+			wg.Add(1)
+			go func(m pkg.InboundMessage) {
```
zhisme, line 139: **Response ordering across sessions on the same channel**

With concurrent goroutines, `ch.Send()` can race — if msg1's handler finishes after msg2's but msg2's `Send` runs first, responses arrive out of order on the channel. The orchestrator's per-session lock helps for same-session ordering, but different sessions on the same channel can interleave `Send` calls.

This is probably acceptable for most channel types, but worth documenting as a behavior change.

internal/orchestrator/orchestrator.go
```diff
@@ -341,9 +359,45 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
+	sm.mu.Unlock()
+}
+
 func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string) (*RunResult, error) {
-	o.mu.Lock()
-	defer o.mu.Unlock()
+	// Global concurrency cap: block until a slot is available (or context is cancelled).
+	select {
+	case o.semaphore <- struct{}{}:
+		defer func() { <-o.semaphore }()
```
zhisme, line 398: **Document lock ordering to prevent future deadlocks**

The acquisition order here is: semaphore → session lock. If any future code path acquires them in the reverse order, you get a deadlock. Consider adding a comment documenting this invariant:

```go
// Lock ordering: semaphore → per-session mutex. Never acquire in reverse.
```

internal/channel/registry.go
```diff
@@ -118,6 +125,10 @@ func (r *Registry) StopAll() {
 func (r *Registry) dispatch(ch pkg.Channel, inbox <-chan pkg.InboundMessage) {
 	defer r.wg.Done()
 
+	sem := make(chan struct{}, r.maxConcurrent)
+	var wg sync.WaitGroup
+	defer wg.Wait() // drain in-flight goroutines on shutdown
```
zhisme: The shutdown flow here is subtle but correct: `r.cancel()` triggers `<-r.ctx.Done()` in the dispatch loop, the loop returns, `defer wg.Wait()` drains child goroutines, then `defer r.wg.Done()` signals the outer WaitGroup. A brief comment clarifying this ordering would help future readers.

zhisme (commented):
## Code Review

The core idea is sound — replacing the global mutex with a semaphore + per-session lock is a good approach. Good test coverage. A few issues to address before merge, one of which is a real bug.

internal/orchestrator/orchestrator.go
```diff
@@ -341,22 +360,64 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
+func (o *Orchestrator) releaseSessionLock(sessionID string, sm *sessionMutex) {
+	o.sessionMuxMu.Lock()
+	sm.refCount--
+	if sm.refCount == 0 {
+		delete(o.sessionMuxes, sessionID)
+	}
+	o.sessionMuxMu.Unlock()
+
+	sm.mu.Unlock()
+}
```
zhisme, line 388: **Race condition concern:** `maybeSummarizeSession` is fired in a background goroutine here. Once `Run` returns, the per-session lock is released, but this goroutine continues reading/writing session messages via `o.sessions` — potentially racing with a subsequent `Run()` on the same session.

Under the old global mutex this was also technically racy, but far less likely since the global lock serialized everything. With concurrent sessions enabled, this becomes a real concern.

Consider having `maybeSummarizeSession` acquire the per-session lock, or document why the session store's internal synchronization is sufficient.

internal/channel/registry.go
```diff
@@ -12,22 +12,29 @@ import (
...
-	channels map[string]pkg.Channel
-	handler  pkg.MessageHandler
+	mu            sync.RWMutex
+	channels      map[string]pkg.Channel
+	handler       pkg.MessageHandler
+	maxConcurrent int // per-channel dispatch concurrency; default 1 (sequential)
 
 	ctx    context.Context
 	cancel context.CancelFunc
 	wg     sync.WaitGroup
```
zhisme, line 23: Both `Registry.dispatch` (this semaphore) and `Orchestrator.Run` (its own `o.semaphore`) enforce `maxConcurrent` — and both are set from the same `max_concurrent_sessions` config value.

This means there are two layers of concurrency limiting competing with each other. If `Orchestrator.Run` already enforces the cap, this semaphore is redundant (and vice versa). Either pick one place to enforce it or document why both are needed (e.g., defense-in-depth).

internal/channel/registry.go
```diff
@@ -12,22 +12,29 @@ import (
...
-		handler:  handler,
-		ctx:      ctx,
-		cancel:   cancel,
+		channels:      make(map[string]pkg.Channel),
+		handler:       handler,
+		maxConcurrent: maxConcurrent,
+		ctx:           ctx,
+		cancel:        cancel,
 	}
 }
```
zhisme, line 36: This comment acknowledges concurrent `ch.Send` calls across sessions. Is thread safety of `Send` guaranteed by the `Channel` interface contract? If not, this PR introduces a new requirement that all channel implementations must have a thread-safe `Send`. Worth documenting on the interface or adding a send mutex here.

zhisme (commented):
## Code Review

Core concurrency design is sound — semaphore for global cap, per-session mutex for ordering, ref-counted cleanup. Good test coverage. A few concerns noted inline.

internal/orchestrator/orchestrator.go
```diff
@@ -341,22 +360,64 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
 func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string) (*RunResult, error) {
-	o.mu.Lock()
-	defer o.mu.Unlock()
+	// Lock ordering: semaphore → per-session mutex. Never acquire in reverse.
+	// Global concurrency cap: block until a slot is available (or context is cancelled).
+	select {
+	case o.semaphore <- struct{}{}:
+		defer func() { <-o.semaphore }()
+	case <-ctx.Done():
+		return nil, ctx.Err()
```
zhisme, line 400: Nit: `pendingMu` is acquired here while the session lock is already held. This ordering is fine as long as no code path ever acquires them in reverse. Consider adding a comment documenting the lock ordering invariant: `semaphore → sessionMux → pendingMu`.

zhisme:
**`RunAction` is unprotected** (`orchestrator.go:959`): `RunAction` calls `executeCall` without acquiring the semaphore or any session lock. If a scheduler/external caller invokes `RunAction` concurrently with a `Run` on the same session, there could be races on shared plugin state. Was this intentional (callers coordinate externally) or should `RunAction` also go through the session lock?

internal/channel/registry.go
```diff
@@ -126,15 +134,21 @@ func (r *Registry) dispatch(ch pkg.Channel, inbox <-chan pkg.InboundMessage) {
...
+			go func(m pkg.InboundMessage) {
+				defer wg.Done()
+
+				sessionKey := pkg.SessionKey(ch.ID(), m.ConversationID, m.ThreadID)
+				resp, err := r.handler(r.ctx, sessionKey, m)
+				if err != nil {
+					log.Printf("handling message on channel %q session %q: %v", ch.ID(), sessionKey, err)
+					return
+				}
+				if err := ch.Send(r.ctx, resp); err != nil {
```
zhisme, line 148: `ch.Send(r.ctx, resp)` is now called from concurrent goroutines (one per inbound message). The `Channel` interface comment in `pkg/channel/types.go` documents this new contract, but `YAMLChannel.Send` (`internal/channel/yaml_channel.go:145`) doesn't satisfy it — it reads shared state (`ch.selfVars`, `ch.config`) via `buildContexts()` and calls `doHTTPCall` / `runHooks` without any synchronization.

Two concurrent `Send` calls on the same `YAMLChannel` can race on map reads/writes, corrupt HTTP request bodies, or interleave chunked sends for different messages.

Suggestion: either serialize `Send` with a mutex inside `YAMLChannel`, or keep the dispatch sequential per-channel (move the goroutine into the orchestrator only) and let the orchestrator's semaphore handle cross-session parallelism.

zhisme (changes requested):
`YAMLChannel.Send` is not safe for concurrent use — see inline comment.

zhisme (approved):
The `selfVars` race in `YAMLChannel` is fixed — `buildContexts()` now snapshots under `RLock`, and `runInit` writes under `Lock`. No remaining major issues. LGTM.

PR #43, opened 2026-03-17
files support

OpakAlex:


zhisme (commented):
## Code Review

**+148 / -12** across 9 files — adds file/image attachment support end-to-end: inbound channels receive base64-encoded file attachments, they flow through the orchestrator to the Anthropic provider as multimodal content blocks, and outbound messages can include files in templates.

---

### Issues & Suggestions

**1. Silently swallowed `json.Marshal` errors** — `internal/provider/anthropic.go`

```go
raw, _ := json.Marshal(m.Content)
// ...
raw, _ := json.Marshal(blocks)
```

If marshaling fails, you get `null` as the message content with no indication anything went wrong. At minimum log the error; better yet return it.

**2. All non-text files are sent as `"type": "image"`** — `internal/provider/anthropic.go`

```go
blocks = append(blocks, anthContentBlock{
    Type: "image",
    ...
})
```

The PR says "files (images, documents, etc.)" but the Anthropic API `"image"` content type only supports images. PDFs use `"document"` type. Sending a PDF as `"type": "image"` will get rejected by the API. You need to dispatch on mime type.

**3. No size limits on file attachments** — `internal/channel/yaml_ws.go`

Any base64 blob in an inbound event gets decoded and held in memory with zero validation. A malicious or buggy client could send a 500MB base64 string and it'd be happily decoded. Add a size cap.

**4. `encodeFilesJSON` silently swallows marshal error** — `internal/channel/yaml_channel.go`

```go
b, _ := json.Marshal(out)
```

Same pattern as #1. At least log it.

**5. File data stored in session history** — `internal/orchestrator/orchestrator.go`

The raw `[]byte` data gets persisted in session state. If sessions are kept in memory, repeated image messages will balloon memory. If sessions are serialized, you're storing potentially large blobs. Worth considering whether to store a reference instead, or at least document the memory implications.

**6. Duplicate data representation** — `pkg/channel/types.go` has `FileAttachment` and `internal/provider/provider.go` has `MessageFile` with nearly identical fields. The conversion in `cmd/opentalon/main.go` just copies `MimeType` + `Data`. Consider whether you actually need two types.

**7. `decodeFileAttachment` accepts empty `name` and `mimeType`** — `internal/channel/yaml_ws.go`

Only `data` being empty triggers a `nil` return. A file with no mime type will break the Anthropic request. Validate required fields.

**8. Named import alias** — `cmd/opentalon/main.go`

```go
pkg "github.com/opentalon/opentalon/pkg/channel"
```

Consider a more descriptive alias than `pkg` (e.g. `chanpkg` or `chantypes`).

---

### What looks good

- Clean threading of files through the handler → orchestrator → provider pipeline
- Variadic `files ...FileAttachment` on the `Runner` interface is backwards-compatible
- Base64 fallback to `RawURLEncoding` in decode is a nice touch
- Outbound template support with both individual `file_N_*` keys and `files_json` gives flexibility

---

### Summary

The biggest concern is **#2** — hardcoding `"type": "image"` for all file types will break for non-image attachments. Fix that, add validation (#3, #7), and handle the marshal errors (#1, #4).

PR #44, opened 2026-03-17
user only actions

OpakAlex:


internal/orchestrator/orchestrator.go
```diff
@@ -548,6 +548,9 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string) (
 			return result, nil
 		}
 
+		for i := range calls {
```
zhisme, line 551: Nit: you could fold this into the existing `for _, call := range calls` loop below to avoid two passes:
```go
for i := range calls {
    calls[i].FromLLM = true
    toolResult := o.executeCall(ctx, calls[i])
    ...
}
```
Not blocking, just fewer iterations.

zhisme (commented):
Good feature — defense-in-depth across three layers (system prompt, planner, execution) is solid. `FromLLM` tagged `yaml:"-"` to prevent spoofing via deserialization is a nice touch. Tests cover all enforcement points well.

Two actionable items:
1. **Log blocked `user_only` attempts** — currently the rejection returns before audit logging runs, which is a security audit gap
2. **Split the protoc-gen-go upgrade** (v1.35.2 → v1.36.11) into its own commit — the `*.pb.go` reformatting is ~80% of the diff and contains no semantic changes beyond the new field

internal/orchestrator/orchestrator.go
```diff
@@ -911,6 +914,12 @@ func (o *Orchestrator) executeCall(ctx context.Context, call ToolCall) ToolResul
 			}
 		}
 	}
+	if call.FromLLM && action != nil && action.UserOnly {
+		return ToolResult{
+			CallID: call.ID,
+			Error:  fmt.Sprintf("action %q can only be invoked by the user, not the LLM", call.Action),
```
zhisme, line 920: **Security gap**: when this rejects a `user_only` call from the LLM, execution returns *before* the audit logging code further down. If `action.AuditLog` is true (like `install_skill`), the blocked attempt is never recorded.

Failed privileged attempts are arguably *more* important to log than successful ones. Consider adding an explicit audit log entry here before returning the error.

internal/orchestrator/orchestrator.go
```diff
@@ -1096,8 +1105,14 @@ func (a *plannerLLMAdapter) Complete(ctx context.Context, req *pipeline.Completi
 func capabilitiesToPlannerInfo(caps []PluginCapability) []pipeline.CapabilityInfo {
 	result := make([]pipeline.CapabilityInfo, len(caps))
 	for i, cap := range caps {
-		actions := make([]pipeline.ActionInfo, len(cap.Actions))
-		for j, a := range cap.Actions {
+		var filteredActions []Action
```
zhisme, line 1108: Minor: `filteredActions` is grown via `append` in a loop. You could pre-count or use `slices.DeleteFunc` to avoid repeated allocations. Not worth blocking over.

pkg/channel/channelpb/channel.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.35.2
+// 	protoc-gen-go v1.36.11
```
zhisme, line 3: The `protoc-gen-go` version bump (v1.35.2 → v1.36.11) reformats all struct layouts and switches the raw descriptor encoding from `[]byte` to string-concat. This is pure noise for reviewers — consider splitting the codegen upgrade into a separate commit so the actual `user_only` logic is easier to review.

internal/orchestrator/types.go
```diff
@@ -25,10 +27,11 @@ type PluginCapability struct {
...
 
 type ToolCall struct {
-	ID     string            `yaml:"id"`
-	Plugin string            `yaml:"plugin"`
-	Action string            `yaml:"action"`
-	Args   map[string]string `yaml:"args,omitempty"`
+	ID      string            `yaml:"id"`
+	Plugin  string            `yaml:"plugin"`
+	Action  string            `yaml:"action"`
+	Args    map[string]string `yaml:"args,omitempty"`
```
zhisme, line 33: Good call on `yaml:"-"` — prevents `FromLLM` from being injected via deserialization. Worth a brief code comment noting this is intentional for security (e.g. `// yaml:"-": must not be deserializable to prevent spoofing`).

PR #53, opened 2026-04-08
support profiles

OpakAlex:


internal/profile/verifier.go
```diff
@@ -0,0 +1,197 @@
...
+
+// Verifier verifies bearer tokens against a WhoAmI HTTP server and caches results.
+type Verifier struct {
+	cfg         VerifierConfig
+	client      *http.Client
+	groupStore  GroupPluginSaver // optional; nil disables auto-save
+	entityStore EntityUpserter   // optional; nil disables entity tracking
+
+	mu    sync.Mutex
+	cache map[[32]byte]cacheEntry
```
zhisme, line 102: **Cache grows unbounded.** The `cache` map is never pruned — entries are only overwritten on the same token hash. In a high-traffic multi-tenant deployment, unique tokens accumulate indefinitely. Consider periodic eviction of expired entries (e.g. a goroutine or lazy sweep on `Verify`) or a bounded LRU.

internal/profile/verifier.go
```diff
@@ -0,0 +1,197 @@
...
+	p, err := v.callServer(ctx, token)
+	if err != nil {
+		return nil, err
+	}
+
+	// Auto-save group→plugin assignments.
+	if v.groupStore != nil && len(p.Plugins) > 0 {
+		if serr := v.groupStore.UpsertGroupPlugins(ctx, p.Group, p.Plugins, "whoami"); serr != nil {
+			// Non-fatal: log but don't fail the request.
+			_ = serr
```
zhisme: **Silent error swallowing.** The comment says "log but don't fail" — but it doesn't actually log. Every other non-fatal failure in the codebase uses `slog.Warn`. Same for the `_ = serr` at line 128. These should at minimum `slog.Warn("auto-save group plugins failed", "error", serr)` for observability.

internal/profile/verifier.go
```diff
@@ -0,0 +1,197 @@
...
+	v.mu.Lock()
+	if e, ok := v.cache[key]; ok && time.Now().Before(e.expiresAt) {
+		v.mu.Unlock()
+		return e.profile, nil
+	}
+	v.mu.Unlock()
+
+	p, err := v.callServer(ctx, token)
+	if err != nil {
+		return nil, err
```
zhisme, line 171: **No negative caching for bad tokens.** Failed verifications are not cached — only successful ones. An attacker spraying unique invalid tokens would cause a flood of HTTP requests to the WhoAmI server. Consider caching negative results for a short TTL (e.g. 10–30s) to rate-limit abuse.

internal/channel/handler.go
```diff
@@ -53,6 +80,15 @@ func NewMessageHandler(
 	}
 }
 
+func errorResponse(msg pkg.InboundMessage, text string) pkg.OutboundMessage {
+	return pkg.OutboundMessage{
+		ConversationID: msg.ConversationID,
+		ThreadID:       msg.ThreadID,
+		Content:        text,
+		Metadata:       msg.Metadata,
```
zhisme: **Potential token leak in error responses.** `msg.Metadata` contains the `profile_token`. Depending on how channels serialize outbound messages, this could leak the bearer token back to the client on error. Consider omitting metadata or stripping sensitive keys (`profile_token`) from error responses.

Same issue exists in the success path at line 78 and the error path at line 64.

internal/state/store/store_test.go
```diff
@@ -79,11 +79,11 @@ func TestMemoryStore_AddScopedAndMemoriesForContext(t *testing.T) {
 		t.Fatalf("MemoriesForContext: %v", err)
 	}
 	if len(list) != 2 {
-		t.Errorf("MemoriesForContext(workflow) for U1: got %d, want 2 (general + U1)", len(list))
+		t.Errorf("MemoriesForContext(workflow) for U1: got %d, want 3 (general + U1)", len(list))
```
zhisme: **Assertion/message mismatch.** The condition checks `!= 2` but the error message says `want 3`. Either the check should be `!= 3` or the message is wrong. This pattern repeats at lines 85-86, 113-114, 126-127, 174-175. These tests pass by accident if the count is still 2, but the error messages are misleading. Looks like a merge artifact.

internal/state/store/group_plugins.go
```diff
@@ -0,0 +1,106 @@
...
+
+// UpsertGroupPlugins upserts the given plugin IDs for groupID with the given source.
+// It will not downgrade an existing row to a lower-priority source.
+func (s *GroupPluginStore) UpsertGroupPlugins(ctx context.Context, groupID string, pluginIDs []string, source string) error {
+	newPri := sourcePriority[source]
+	now := time.Now().UTC().Format(time.RFC3339)
+	for _, pid := range pluginIDs {
+		// Check existing priority.
+		var existingSource string
+		err := s.db.QueryRowContext(ctx,
```
zhisme: **N+1 queries per upsert.** Each plugin ID requires a SELECT + INSERT/UPDATE, so upserting 10 plugins = 20 queries. This runs on every WhoAmI response that includes plugins. Consider wrapping in a transaction and/or using a single `INSERT ... ON CONFLICT` with the priority check expressed in SQL (e.g. via a CASE on the source column).

Not blocking, but worth addressing if WhoAmI responses frequently include plugin lists.

internal/orchestrator/orchestrator.go
```diff
@@ -891,14 +948,30 @@ func (o *Orchestrator) executeCall(ctx context.Context, call ToolCall) ToolResul
...
 				action = &cap.Actions[i]
 				break
 			}
 		}
 	}
+
+	// Defense-in-depth: block restricted plugins when a profile is active but the plugin is not allowed.
+	// This mirrors the filtering done in buildSystemPrompt; protects against crafted tool calls.
+	if len(capForCheck.AllowedGroups) > 0 {
+		allowedPlugins := o.resolveAllowedPlugins(ctx)
```
zhisme, line 969: **`resolveAllowedPlugins` called twice per restricted tool call.** This was already resolved at line 816 in `buildSystemPrompt`. Here in `executeCall` it hits the DB again. For the normal path this is fine (defense-in-depth only fires on crafted calls), but consider passing the allowed set through context to avoid the redundant lookup.

cmd/opentalon/main.go
```diff
@@ -87,23 +88,34 @@ func main() {
...
+			groupPluginStore = store.NewGroupPluginStore(db)
+			usageStore = store.NewUsageStore(db)
+			entityStore = store.NewEntityStore(db)
+			// Seed static group→plugin assignments from config (source="config"; does not overwrite whoami/admin).
+			seedGroupPlugins(context.Background(), groupPluginStore, cfg.Profiles.Groups)
 		}
 	} else {
 		memory, sessions = newInMemoryState()
 	}
+	_ = stateDB // used indirectly via stores
```
zhisme: Code smell — `stateDB` only exists because the stores are declared before the `if` block. The stores already hold a reference to the underlying `*sql.DB`. Consider declaring the stores inside the `if` block (they're only assigned there anyway) and dropping `stateDB` entirely.

internal/state/store/migrations/003_usage_model.sql
```diff
@@ -0,0 +1,5 @@
+-- Add model tracking and cost fields to profile_usage.
```
zhisme: Since this feature hasn't shipped yet, consider squashing migration 002 and 003 into a single migration. Having 002 create the `profile_usage` table and 003 immediately `ALTER TABLE` it adds unnecessary upgrade steps for anyone running these migrations fresh.

zhisme (changes requested):
## Review: support profiles

Solid multi-tenancy feature with good architecture — clean package separation, defense-in-depth plugin filtering, fail-closed design, and good test coverage. A few things should be addressed before merge (see inline comments).

internal/state/store/store_test.go
zhisme: **Bug:** The condition checks `!= 2` but the error message says "want 3". Same pattern repeats below for `U2`, `SessionStore_PersistAndGet`, and `SetSummaryRoundTrip`. Either the assertion or the message is wrong — if the expected count changed to 3 with migration 002, the condition should be `!= 3`; if it's still 2, revert the message.

internal/orchestrator/orchestrator.go
```diff
@@ -1070,6 +1147,64 @@ func formatToolResultMessage(result ToolResult) string {
...
+// The result is cached in ctx by Run; within a single Run call this never hits the DB twice.
+func (o *Orchestrator) resolveAllowedPlugins(ctx context.Context) map[string]bool {
+	if cached, ok := ctx.Value(allowedPluginsKey{}).(cachedAllowedPlugins); ok {
+		return cached.m
+	}
+	if o.groupPluginLookup == nil {
+		return nil
+	}
+	p := profile.FromContext(ctx)
+	if p == nil || p.Group == "" {
```
zhisme: **Security concern:** When a profile exists but has an empty `group`, this returns `nil` (= no restrictions), which means the entity gets unrestricted access to **all** plugins, including those with `AllowedGroups` set.

If the WhoAmI server omits `group` for some entities, those entities silently get more access than entities that *do* have a group. Consider returning `map[string]bool{}` (empty map = block all restricted plugins) when a profile is present but `Group` is empty. That way profiles without a group only see unrestricted plugins, which is the safer default.

internal/profile/verifier.go
```diff
@@ -0,0 +1,227 @@
...
+
+// Verifier verifies bearer tokens against a WhoAmI HTTP server and caches results.
+type Verifier struct {
+	cfg         VerifierConfig
+	client      *http.Client
+	groupStore  GroupPluginSaver // optional; nil disables auto-save
+	entityStore EntityUpserter   // optional; nil disables entity tracking
+
+	mu    sync.Mutex
+	cache map[[32]byte]cacheEntry
```
zhisme, line 102: **Unbounded cache:** This map has no max-size limit. A flood of unique tokens (even invalid ones — negative cache TTL is 15s) can grow this indefinitely. The inline eviction only removes expired entries, not excess ones.

Consider either:
- A hard cap (e.g., evict oldest entries when exceeding N), or
- An LRU cache with a max size (e.g., `groupcache/lru` or a simple wrapper)

internal/profile/verifier.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	v.mu.Unlock()
+
+	p, err := v.callServer(ctx, token)
+	if err != nil {
+		var rejected rejectedError
+		if errors.As(err, &rejected) {
+			v.mu.Lock()
+			now := time.Now()
+			for k, e := range v.cache {
```
zhisme: **Perf nit:** This full-map iteration under `v.mu` runs on every cache miss (both positive and negative paths at line 157). With high token cardinality this becomes O(n) lock contention. A periodic background goroutine with a ticker would keep the hot path fast.

internal/profile/verifier.go
```diff
@@ -0,0 +1,227 @@
...
+	}, nil
+}
+
+func jsonString(raw json.RawMessage) string {
+	if raw == nil {
+		return ""
+	}
+	var s string
+	if err := json.Unmarshal(raw, &s); err != nil {
+		return strings.Trim(string(raw), `"`)
```
zhisme: **Fragile fallback:** If the WhoAmI server returns `entity_id` as a number (`"entity_id": 123`), `json.Unmarshal` into `string` fails and this fallback returns `"123"` — which silently works but is a type confusion. Worse, if the value is an object or array you get garbage.

Safer to return `""` on unmarshal failure and let the caller's "missing field" validation catch it:
```go
if err := json.Unmarshal(raw, &s); err != nil {
    return ""
}
```

internal/state/store/group_plugins.go
```diff
@@ -0,0 +1,110 @@
...
+	_, err := s.db.ExecContext(ctx,
+		`DELETE FROM group_plugins WHERE group_id = ? AND plugin_id = ?`, groupID, pluginID)
+	if err != nil {
+		return fmt.Errorf("group_plugins: revoke %s/%s: %w", groupID, pluginID, err)
+	}
+	return nil
+}
+
+// ListGroups returns all group IDs that have at least one plugin assigned.
+func (s *GroupPluginStore) ListGroups(ctx context.Context) ([]string, error) {
```
zhisme: **Unused code:** `ListGroups` has no callers. If it's for a future feature, I'd remove it until there's a consumer — dead code accumulates silently and the function is trivial to re-add.

internal/auth/auth_test.go
```diff
@@ -255,6 +255,7 @@ func TestStoreAddAndGet(t *testing.T) {
 	p := store.Get("openai:default")
 	if p == nil {
 		t.Fatal("Get returned nil")
+		return
```
zhisme: Nit: `t.Fatal` already stops the test goroutine via `runtime.Goexit()`, so this `return` is dead code. Same at lines 321 and 332. Harmless, but if this is appeasing a linter it's worth noting in the commit message.

zhisme (commented):
Good feature — architecture is solid and the priority-based group→plugin system is well-designed. A few issues to address, mostly around the test assertion bug and a potential security gap with empty groups.

PR #63, opened 2026-04-10
send channel user to auth service

OpakAlex:


internal/profile/verifier.go
```diff
@@ -29,27 +30,30 @@ type EntityUpserter interface {
...
 	}
 	if c.TokenHeader == "" {
+		// Apply defaults only when the header is unconfigured.
+		// If a custom TokenHeader is set, TokenPrefix defaults to "" (no prefix).
 		c.TokenHeader = "Authorization"
-	}
-	if c.TokenPrefix == "" {
-		c.TokenPrefix = "Bearer "
+		if c.TokenPrefix == "" {
+			c.TokenPrefix = "Bearer "
```
zhisme, line 56: Silent breaking change: previously any custom `TokenHeader` still got `"Bearer "` prepended by default. After this PR, `token_header: X-Custom` with no explicit `token_prefix` sends `X-Custom: <token>` instead of `X-Custom: Bearer <token>`. Probably the right default, but existing configs will flip on upgrade with no warning.

At minimum: call it out in the PR description / release notes. Better: add a regression test pinning both branches (default header keeps `Bearer `, custom header defaults to no prefix).

internal/profile/verifier.go
```diff
@@ -220,6 +224,9 @@ func (v *Verifier) callServer(ctx context.Context, token string) (*Profile, erro
 		return nil, fmt.Errorf("%w: build request: %v", ErrAuthFailed, err)
 	}
 	req.Header.Set(v.cfg.TokenHeader, v.cfg.TokenPrefix+token)
+	for k, val := range v.cfg.ExtraHeaders {
+		req.Header.Set(k, os.ExpandEnv(val))
```
zhisme: Two issues with expanding per-call:

1. `os.ExpandEnv` silently replaces unknown `${FOO}` with empty string. If `WHOAMI_SECRET` is unset or typo'd, this header goes out empty and every WhoAmI call 401s with no signal from our side.
2. The cache key is the token only, so a mid-run env change wouldn't invalidate entries — stale secret stays cached.

Suggest expanding once in `setDefaults` and either failing fast or logging a warning when a `${VAR}` resolves empty. That also makes `ExtraHeaders` immutable for the verifier's lifetime, which matches how secrets are typically loaded.

internal/profile/verifier.go
```diff
@@ -220,6 +224,9 @@ func (v *Verifier) callServer(ctx context.Context, token string) (*Profile, erro
...
+	}
```
zhisme, line 250: Header-collision footgun: `TokenHeader` is `Set` first, then this loop `Set`s ExtraHeaders, so a misconfigured `extra_headers` entry named the same as `token_header` (both default to `Authorization`) silently overwrites the token.

Cheap guard: skip keys whose canonical form equals `textproto.CanonicalMIMEHeaderKey(v.cfg.TokenHeader)`, and log once at startup if that happens.

internal/config/config.go
```diff
@@ -42,10 +42,11 @@ type WhoAmIConfig struct {
...
 	CacheTTL      string `yaml:"cache_ttl"`       // Go duration; default "60s"
-	EntityIDField string `yaml:"entity_id_field"` // JSON field in response; default "entity_id"
-	GroupField    string `yaml:"group_field"`     // JSON field in response; default "group"
-	PluginsField  string `yaml:"plugins_field"`   // optional JSON field for plugin list; default "plugins"
-	ModelField    string `yaml:"model_field"`     // optional JSON field for model override; default "model"
+	EntityIDField string            `yaml:"entity_id_field"`        // JSON field in response; default "entity_id"
+	GroupField    string            `yaml:"group_field"`            // JSON field in response; default "group"
+	PluginsField  string            `yaml:"plugins_field"`          // optional JSON field for plugin list; default "plugins"
+	ModelField    string            `yaml:"model_field"`            // optional JSON field for model override; default "model"
+	ExtraHeaders  map[string]string `yaml:"extra_headers,omitempty"` // static headers added to every WhoAmI call; values support ${ENV_VAR}
```
zhisme, line 63: Struct alignment got wider because of the new `map[string]string` column — existing string fields above shifted too. Looks gofmt-clean in the diff, just double-check `gofmt -d` locally so nothing weird sneaks in.

internal/profile/verifier_test.go
```diff
@@ -91,6 +91,66 @@ func TestVerifier_AuthFailed_MissingEntityID(t *testing.T) {
...
+		}
+		_ = json.NewEncoder(w).Encode(map[string]interface{}{"entity_id": "u1"})
+	}))
+	defer srv.Close()
+
+	v := NewVerifier(VerifierConfig{
+		URL:          srv.URL,
+		TokenHeader:  "X-User-ID",
+		TokenPrefix:  "",
+		ExtraHeaders: map[string]string{"X-Security-Token": "${TEST_WHOAMI_SECRET}"},
```
zhisme, line 147: Worth tightening: `err == nil` passes for any error, including transport errors. Prefer `errors.Is(err, ErrAuthFailed)` so a future refactor that breaks the error wrapping gets caught here.

Also missing: a test for the unset-env-var case (expansion to empty string). That's the silent failure mode from the `os.ExpandEnv` comment above — would be the natural place to pin whatever behavior we decide on.

docs/profiles.md
```diff
@@ -67,8 +67,52 @@ profiles:
     group_field: group            # JSON key for group; default "group"
     plugins_field: plugins        # JSON key for plugin list; default "plugins"
     model_field: model            # JSON key for model override; default "model"
+    extra_headers: {}             # static headers added to every WhoAmI call (see below)
```
zhisme, line 70: nit: doc is clear and the warning about platform user IDs not being secret is good. One thing worth adding: the shared secret assumes TLS between OpenTalon and the WhoAmI server — it's replay-vulnerable otherwise. One sentence is enough.

zhisme (commented):
Overall: feature is small and useful. Two blockers below: the silent `TokenPrefix` default change, and the per-call `os.ExpandEnv` that swallows missing env vars. Header-collision guard and tighter tests are nice-to-haves.

internal/bootstrap/provider.go
```diff
@@ -0,0 +1,195 @@
...
+		req.Header.Set(p.cfg.TokenHeader, p.cfg.TokenPrefix+p.cfg.Token)
+	}
+
+	resp, err := p.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("bootstrap: fetch %s: %w", p.cfg.URL, err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
zhisme: 5xx is treated as terminal here, same as 4xx. Combined with the `Fetch` docstring at line 73 ("retries up to cfg.Retries times with linear backoff... 4xx HTTP responses are treated as terminal"), this is misleading: retries only cover network errors (connection refused, DNS, TLS). A transient 503 from the bootstrap server won't be retried.

The author's own test comment in `provider_test.go` acknowledges the confusion:
```
// 5xx is not in the terminal-error path ...
// Actually our terminalError wraps 4xx. 5xx falls through as a non-terminal error? Let me check...
// No: doFetch returns terminalError for ANY non-2xx ...
// So 5xx IS terminal too in the current impl. Let's test actual network retry instead.
```
That comment should not ship — it's the author debugging the semantics in prose. Pick one: either (a) split 4xx/5xx so 5xx retries as "transient", which is the canonical retry target, or (b) update the docstring to say "all non-2xx responses are terminal; only network errors are retried" and delete the confused comment.

internal/bootstrap/provider.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, &terminalError{fmt.Errorf("bootstrap: parse body: %w", err)}
+	}
+	return &out, nil
+}
+
+// Merge merges the remote Response into the static config and returns a new Config.
+// Static config entries always win on key conflicts — remote entries are purely additive.
+// GroupPlugins from the remote response is not merged into the config struct; callers
+// handle seeding it to the DB separately via the returned Response.GroupPlugins field.
+// Merge merges the remote Response into the static config and returns a new Config.
```
zhisme, line 132: Duplicate docstring — the `// Merge merges the remote Response into the static config...` paragraph appears twice (lines 129-132 and lines 133-136) before the function declaration. Accidental paste. Keep the second version (mentions "The static config is never mutated") and drop the first.

internal/bootstrap/provider.go
```diff
@@ -0,0 +1,195 @@
...
+// Static config entries always win on key conflicts — remote entries are purely additive.
+// The static config is never mutated. GroupPlugins from the response is not merged into
+// the returned Config; callers seed it to the DB separately.
+func Merge(static *config.Config, remote *Response) *config.Config {
+	if remote == nil {
+		return static
+	}
+	merged := *static
```
zhisme, line 143: `merged := *static` is a shallow copy. `Channels` and `Plugins` are rebuilt below so those are safe, but every other field on `Config` (other maps, slices, pointers like `LuaConfig`) remains aliased to the static config. Right now callers don't mutate those after Merge, but this is a fragile contract — the `TestMerge_DoesNotMutateStatic` test only checks `Channels` and `Plugins`, so regressions in other fields would go unnoticed.

Either document the contract loudly ("only Channels/Plugins are deep-copied — do not mutate any other field of the returned Config") or deep-copy the remaining top-level maps too.

zhisme (commented):
Round 2.

**WhoAmI / extra_headers** — all six items from the previous round look cleanly addressed:
- ✅ `os.ExpandEnv` moved to `setDefaults`, runs once at construction
- ✅ Empty-after-expansion now logs a warning
- ✅ Header collision with `TokenHeader` is detected via canonical comparison
- ✅ `errors.Is(err, ErrAuthFailed)` in tests
- ✅ New tests for expansion-at-construction, unset env var, and token-header collision
- ✅ TLS note added to docs

One nit: the bootstrap package has a dedicated `TestNew_TokenPrefixDefaults` regression test for the default-change, but the profile verifier (where the original regression would have happened) doesn't. Not a blocker — the `ExtraHeaders` tests exercise the `TokenHeader != ""` branch indirectly.

**Bootstrap feature — scope concern.** The PR title is "send channel user to auth service" but this now adds a full remote bootstrap fetcher (~550 LoC: new package, main.go integration, config plumbing, group_plugins priority rework). This is an orthogonal feature — unrelated to the WhoAmI extra_headers work — and reviewing them together makes both harder. If it isn't already merged downstream somewhere, I'd prefer this land as a separate PR. If you'd rather keep it bundled, rename the PR title so reviewers know what they're signing up for.

The bootstrap code itself has a few issues flagged inline — the 5xx/retry semantics and the group_plugins priority tie are the ones worth fixing before merge.

internal/state/store/group_plugins.go
```diff
@@ -54,8 +54,8 @@ VALUES (?, ?, ?, ?, ?)
 ON CONFLICT (group_id, plugin_id) DO UPDATE SET
     source     = excluded.source,
     updated_at = excluded.updated_at
-WHERE CASE excluded.source WHEN 'config' THEN 1 WHEN 'whoami' THEN 2 WHEN 'admin' THEN 3 ELSE 0 END
-   >= CASE source           WHEN 'config' THEN 1 WHEN 'whoami' THEN 2 WHEN 'admin' THEN 3 ELSE 0 END`
+WHERE CASE excluded.source WHEN 'config' THEN 1 WHEN 'bootstrap' THEN 1 WHEN 'whoami' THEN 2 WHEN 'admin' THEN 3 ELSE 0 END
+   >= CASE source           WHEN 'config' THEN 1 WHEN 'bootstrap' THEN 1 WHEN 'whoami' THEN 2 WHEN 'admin' THEN 3 ELSE 0 END`
```
zhisme: `bootstrap` and `config` are both assigned priority `1`, and the `WHERE` clause uses `>=`. Combined with the seeding order in `main.go` (`seedGroupPlugins` first → `seedBootstrapGroupPlugins` second), **bootstrap overwrites config on a tie**. But the bootstrap docs and Merge function both say "static config wins on key conflicts" — so `group_plugins` has the opposite precedence from `channels` and `plugins`. Subtle footgun.

Fixes, pick one:
1. Swap the seeding order in main.go — bootstrap first, then config. With `>=`, later same-priority upserts win, so config would take precedence. This matches the documented contract.
2. Give `bootstrap` priority `0` (below `config`), so it can never overwrite config-seeded entries.

Option 1 is smaller. Either way, add a test that seeds both and asserts which wins.

internal/config/config.go
```diff
@@ -413,6 +429,13 @@ func expandEnvInPlugins(cfg *Config) {
 	}
 }
 
+func expandEnvInBootstrap(cfg *Config) {
+	cfg.Bootstrap.URL = expandEnv(cfg.Bootstrap.URL)
+	cfg.Bootstrap.Token = expandEnv(cfg.Bootstrap.Token)
+	cfg.Bootstrap.TokenHeader = expandEnv(cfg.Bootstrap.TokenHeader)
+	cfg.Bootstrap.TokenPrefix = expandEnv(cfg.Bootstrap.TokenPrefix)
```
zhisme, line 436: nit: expands URL, Token, TokenHeader, TokenPrefix — but not `Timeout`. Low impact (who puts env vars in a duration?), just inconsistent with how `expandEnvInPlugins` / `expandEnvInChannels` expand everything they can. Either add it or leave a `// Timeout intentionally not expanded` note.

internal/bootstrap/provider.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	return &Provider{
+		cfg:    cfg,
+		client: &http.Client{Timeout: timeout},
+	}
+}
+
+// Fetch calls the remote endpoint and returns the parsed Response.
+// On transient errors it retries up to cfg.Retries times with linear backoff.
+// 4xx HTTP responses are treated as terminal (no retry).
```
zhisme: nit: the comment says "4xx HTTP responses are treated as terminal" but the implementation at line 113 treats **all** non-2xx as terminal. Either fix the docstring to match the code, or fix the code to match the docstring (see comment there). These should agree so the next person reading this doesn't have to run the tests to figure out what retries actually do.

PR #64, opened 2026-04-10
Multi database dialect abstraction

OpakAlex:


internal/state/store/dialect.go
```diff
@@ -0,0 +1,105 @@
...
+
+// BeginExclusive starts a serialised write transaction.
+// The returned cleanup func must be deferred by the caller to release connection resources.
+//
+//	SQLite:   dedicated Conn + BEGIN IMMEDIATE (blocks concurrent writers at start)
+//	Postgres: db.BeginTx with LevelSerializable
+func (d Dialect) BeginExclusive(ctx context.Context, db *sql.DB) (ExclusiveTx, func(), error) {
+	if d.name == "postgres" {
+		tx, err := db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelSerializable})
+		if err != nil {
```
zhisme, line 75: Correctness bug: `LevelSerializable` is **not** equivalent to SQLite `BEGIN IMMEDIATE`.

- SQLite `BEGIN IMMEDIATE` acquires a write lock at begin time, so a second writer **blocks** until the first commits and then reads the committed state. `AddMessage`'s read-modify-write relies on this.
- Postgres `SERIALIZABLE` is optimistic — it does **not** block. Two concurrent `AddMessage` calls on the same session can both proceed, both read the same `messages`, and the second `Commit()` fails with `40001 serialization_failure`. `AddMessage` returns that error and the append is lost from the caller's perspective.

Under any real concurrency on Postgres you get lost messages + user-visible errors.

Fix options:
1. **Preferred:** do a normal read-committed tx and `SELECT messages FROM sessions WHERE id = ? FOR UPDATE` — this is the direct row-level equivalent of `BEGIN IMMEDIATE` and needs no retry loop. Push the `FOR UPDATE` fragment through the dialect (empty string on sqlite).
2. Keep `SERIALIZABLE` and add a retry loop on `pq.Error.Code == "40001"` around the whole read-modify-write in `AddMessage`.

Option 1 is cleaner and I'd drop `ExclusiveTx` / `sqliteConnTx` entirely in favour of it, but if you want the smallest diff, option 2 is acceptable as long as the retry is actually there.

zhisme (commented):
Refactor shape is good — narrow `Dialect` boundary, mechanical per-store rebinds, `*DB` pass-through. Two blockers before merge: (1) `AddMessage` is not correct under concurrent writers on Postgres — see the inline on dialect.go:65; (2) zero test coverage for the Postgres path, so `Rebind` / `TagMatch` / migrations / concurrency are all untested. Rest is cleanup — details inline.

internal/state/store/dialect.go
```diff
@@ -0,0 +1,105 @@
...
+	PostgresDialect = Dialect{"postgres"}
+)
+
+// Rebind converts ? placeholders to $1, $2, … for PostgreSQL; no-op for SQLite.
+func (d Dialect) Rebind(query string) string {
+	if d.name != "postgres" {
+		return query
+	}
+	var b strings.Builder
+	n := 0
```
zhisme, line 25: `Rebind` walks raw bytes and rewrites every `?`, including ones inside quoted string literals. None of the current queries trigger this, but a future `WHERE name = '?'` or a LIKE pattern containing `?` will silently break on Postgres only.

Cheapest fix: add a comment here saying "do not use `?` inside SQL string literals — this rewriter is not quote-aware". If you want to be safer, skip content between unescaped single quotes like `sqlx.Rebind` does. Leaning toward the comment — keeping it dumb is fine if the rule is written down next to the implementation.

internal/state/store/session.go
```diff
@@ -83,22 +83,19 @@ func (s *SessionStore) Create(id string) *state.Session {
 }
 
 // AddMessage appends a message and persists. If maxMessages > 0, trims to last maxMessages.
-// Uses a reserved connection and BEGIN IMMEDIATE so the second writer blocks until the first commits.
+// Uses an exclusive transaction so concurrent writers are serialised.
 func (s *SessionStore) AddMessage(id string, msg provider.Message) error {
```
zhisme, line 87: This function's correctness depends entirely on `BeginExclusive` actually serialising writers — see the comment on `dialect.go:65`. On the Postgres path it doesn't, and there's no retry here, so concurrent `AddMessage` calls on the same session will drop messages.

If you go with `SELECT ... FOR UPDATE` in the dialect, the fix here is trivial (add the fragment before the select). If you go with serialization-failure retry, wrap the whole body from `BeginExclusive` through `Commit` in a bounded retry loop.

internal/state/store/db.go
```diff
@@ -83,23 +127,23 @@ func (d *DB) runMigrations() error {
...
+		sqlStr, err := migrationSQL(name)
 		if err != nil {
 			return fmt.Errorf("migration %s: %w", name, err)
 		}
 		tx, err := d.db.Begin()
 		if err != nil {
 			return fmt.Errorf("migration %s: begin: %w", name, err)
 		}
-		if _, err := tx.Exec(sql); err != nil {
+		if _, err := tx.Exec(sqlStr); err != nil {
```
zhisme, line 138: Migrations run unchanged against both backends. The two existing files happen to be portable (`TEXT`/`INTEGER`/`REAL`, `CREATE TABLE IF NOT EXISTS`, `ON CONFLICT` with column target all work in both), but:

1. There's no test that actually opens a Postgres instance, so nothing catches the day someone adds `AUTOINCREMENT`, `WITHOUT ROWID`, or uses `json_each` in a migration.
2. `runMigrations` wraps each file in a single tx — fine on both engines, but on Postgres a failure aborts the tx and any subsequent statement in the same migration file will fail with `current transaction is aborted`. Worth knowing.

Minimum viable: add a `//go:build postgres` integration test that reads `DATABASE_URL` (or spins up PG via testcontainers) and runs `TestOpenAndMigrations` + the concurrent `AddMessage` scenario against it. Without it, "multi-dialect" is aspirational — the refactor compiles but nothing proves it works.

go.mod
```diff
@@ -15,6 +15,7 @@ require github.com/google/uuid v1.6.0
 
 require (
 	github.com/dustin/go-humanize v1.0.1 // indirect
+	github.com/lib/pq v1.12.3 // indirect
```
zhisme: Two things:

1. `lib/pq` is marked `// indirect` but `internal/state/store/db.go` has `_ "github.com/lib/pq"` — that makes it a direct dependency. Run `go mod tidy`; the fact that it's still indirect suggests the module graph wasn't cleaned up.

2. Soft push: `lib/pq` has been in maintenance mode for years, upstream README recommends `jackc/pgx`. For a new PG integration in 2026 `pgx/stdlib` is the better default — actively maintained, better perf, proper type support, and crucially `pgconn.PgError` gives you structured SQLSTATE codes, which you'll need if you go with the retry-on-serialization-failure fix for `AddMessage`. Push back if there's a reason you specifically want lib/pq.

PR #65, opened 2026-04-12
enable cluster mode

OpakAlex:


internal/channel/registry.go
```diff
@@ -35,6 +45,13 @@ func NewRegistry(handler pkg.MessageHandler) *Registry {
 	}
 }
 
+// SetDeduplicator attaches a Redis-backed deduplicator to the registry.
+// Must be called before any channels are registered.
+func (r *Registry) SetDeduplicator(d MessageDeduplicator, ttl time.Duration) {
+	r.dedup = d
```
zhisme, line 53: Data race: `r.dedup` / `r.dedupTTL` are written here without holding `r.mu`, while `dispatch` reads `r.dedup` from per-message goroutines. The doc comment says "must be called before any channels are registered" but nothing enforces it. Either take `r.mu.Lock()` here (and use `RLock` at the read site), or `panic` if `len(r.channels) > 0`.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,201 @@
...
+
+	"github.com/opentalon/opentalon/internal/actor"
+	"github.com/opentalon/opentalon/internal/profile"
+)
+
+const (
+	execStream      = "opentalon:plugin-exec"
+	replyStreamFmt  = "opentalon:plugin-exec-reply:%s"
+	consumerGroup   = "opentalon-dispatcher"
+	consumerName    = "dispatcher"
```
zhisme: `consumerName` is a constant, so every pod registers as the same consumer in the group. Work distribution still works, but if a pod dies mid-handle, its pending entries are owned by "dispatcher" forever — no other pod auto-claims them. Use `os.Hostname()` (or a uuid) per pod, and add a periodic `XAUTOCLAIM` sweeper for stale pending entries.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,201 @@
...
+			continue
+		}
+		if err != nil {
+			slog.Warn("plugin exec: xreadgroup error", "error", err)
+			time.Sleep(time.Second)
+			continue
+		}
+		for _, stream := range streams {
+			for _, msg := range stream.Messages {
+				go d.handle(ctx, msg)
```
zhisme: Two issues:

1. **Unbounded goroutines.** A burst on the stream spawns a goroutine per message with no semaphore. Bound concurrency (e.g. a worker pool sized from config).
2. **Shutdown races ack.** `Start` has no `sync.WaitGroup` for spawned `handle`s. When `ctx` is cancelled, `Start` returns, but handlers are still running and their `defer XAck(ctx, ...)` runs against a cancelled context → work executes but ack fails → redelivery on restart. Track handlers in a WaitGroup and `Wait()` before returning from `Start`.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,201 @@
...
+
+	reqCtx := ctx
+	if req.EntityID != "" || req.GroupID != "" {
+		p := &profile.Profile{
+			EntityID:  req.EntityID,
+			Group:     req.GroupID,
+			ChannelID: req.ChannelID,
+		}
+		reqCtx = profile.WithProfile(reqCtx, p)
+		reqCtx = actor.WithActor(reqCtx, req.EntityID)
```
zhisme, line 261: **Authorization gap.** The dispatcher reconstructs `actor` / `profile` purely from fields in the stream entry, then runs the action with full ToolRegistry privileges. Anyone with `XADD` on `opentalon:plugin-exec` can execute any plugin action impersonating any `entity_id` / `group_id`. In cluster mode this Redis is shared with the dedup path, so blast radius is every plugin.

The PR describes this as "trusted plugins" but nothing in the protocol enforces trust. At minimum:
- Loudly document that this Redis must be isolated (not shared with anything else, network-restricted).
- Better: HMAC-sign each request with a shared secret from config; reject unsigned entries.

zhisme (commented):
Inline notes below. Full writeup already in chat — most important is the trust model on the exec stream + auto-exposed plugin HTTP.

internal/plugin/manager.go
```diff
@@ -152,6 +153,16 @@ func (m *Manager) Load(ctx context.Context, entry PluginEntry) error {
 		m.watchProcess(ctx, entry.Name, proc)
 	}
 
+	// If the plugin declared an HTTP address in its handshake, register a reverse proxy
+	// on the shared webhook server so its REST API is accessible at /{plugin-name}/*.
+	if httpAddr := client.HTTPAddr(); httpAddr != "" {
+		if err := channel.RegisterReverseProxy(0, entry.Name, httpAddr); err != nil {
```
zhisme, line 164: Any plugin that sets `OPENTALON_HTTP_PORT` automatically gets `/{plugin-name}/*` exposed on the webhook server with no auth. The webhook server is presumably internet-facing (Slack webhooks), so a misconfigured plugin happily publishes its admin API.

Require an explicit host-side opt-in in plugin config (e.g. `expose_http: true`) before registering the proxy. Plugin-declared handshake field alone isn't enough — the operator should have the final say.

internal/channel/webhook_server.go
```diff
@@ -66,6 +69,27 @@ func (s *WebhookServer) register(port int, path string, handler http.HandlerFunc
...
+func RegisterReverseProxy(port int, prefix, targetAddr string) error {
+	port = webhookPortOrDefault(port)
+	target, err := url.Parse("http://" + targetAddr)
+	if err != nil {
+		return fmt.Errorf("invalid proxy target %q: %w", targetAddr, err)
+	}
+	proxy := httputil.NewSingleHostReverseProxy(target)
+	pattern := "/" + strings.Trim(prefix, "/") + "/"
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r.URL.Path = strings.TrimPrefix(r.URL.Path, "/"+strings.Trim(prefix, "/"))
```
zhisme: Only `r.URL.Path` is rewritten; `r.URL.RawPath` (used for percent-encoded paths) is left alone. `httputil.NewSingleHostReverseProxy` will forward the original RawPath if set, producing a mismatch for URLs containing `%2F`, spaces, etc. Either clear `RawPath` as well, or use `http.StripPrefix` which handles both.

cmd/opentalon/main.go
```diff
@@ -407,6 +408,34 @@ func main() {
...
+		var dedupClient dedup.Deduplicator
+		var dedupErr error
+		if len(cfg.Cluster.Sentinels) > 0 {
+			dedupClient, dedupErr = dedup.NewSentinel(cfg.Cluster.MasterName, cfg.Cluster.Sentinels, cfg.Cluster.Password, cfg.Cluster.SentinelPassword)
+		} else {
+			dedupClient, dedupErr = dedup.NewStandalone(cfg.Cluster.RedisURL)
+		}
+		if dedupErr != nil {
+			fmt.Fprintf(os.Stderr, "Error connecting to Redis for deduplication: %v\n", dedupErr)
+			return
```
zhisme: `return` from `main` exits 0. A k8s pod that can't reach Redis at startup silently "succeeds" and restarts… wait, no — it doesn't restart, because exit 0 is treated as success under `restartPolicy: Always` (yes it does restart, but with backoff behaving like a normal completion rather than a crashloop). Either way, `os.Exit(1)` / `log.Fatal` is what you want here so operators see a real failure. Also applies to the other `fmt.Fprintf(os.Stderr, ...); return` in this file, but those are pre-existing.

internal/channel/registry_dedup_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	ts := time.Now()
+	msg := newMsg("slack", "C123", ts)
+
+	// Send the same message twice (simulates two pods both receiving it).
+	ch.inbox <- msg
+	ch.inbox <- msg
+
+	// Give dispatch goroutines time to process.
+	time.Sleep(100 * time.Millisecond)
```
zhisme: `time.Sleep(100ms)` to wait for dispatch goroutines will flake on busy CI. Have the handler `Done()` a `sync.WaitGroup` the test provides, or wait on `handled` with a deadline (`for deadline := time.Now().Add(2*time.Second); handled.Load() < expected && time.Now().Before(deadline); ...`). Same pattern at lines 115, 141, 168, 195.

internal/dedup/redis.go
```diff
@@ -0,0 +1,73 @@
...
+	if err := client.Ping(ctx).Err(); err != nil {
+		_ = client.Close()
+		return nil, fmt.Errorf("dedup: connecting to redis: %w", err)
+	}
+	return &redisDedup{client: client}, nil
+}
+
+// TryAcquire sets key with NX+EX. Returns true if this call created the key (won the lock).
+func (d *redisDedup) TryAcquire(ctx context.Context, key string, ttl time.Duration) (bool, error) {
+	return d.client.SetNX(ctx, key, 1, ttl).Result()
```
zhisme: Minor: storing the literal `1` loses the one piece of info that would be useful when debugging "why did this key not expire / who claimed it" — the pod identity. Consider storing `os.Hostname()` (or a pod ID) as the value.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,280 @@
...
+			continue
+		}
+		if err != nil {
+			slog.Warn("plugin exec: xreadgroup error", "error", err)
+			time.Sleep(time.Second)
+			continue
+		}
+		for _, stream := range streams {
+			for _, msg := range stream.Messages {
+				d.sem <- struct{}{} // acquire semaphore; blocks if at capacity
```
zhisme: **Blocker: shutdown hang.** Unguarded send on `d.sem`. If the semaphore is full when `ctx` is cancelled, this blocks forever — `Start()` never returns and the `defer wg.Wait()` never fires. Same pattern at line 182 in `autoclaimLoop`.

```go
select {
case d.sem <- struct{}{}:
case <-ctx.Done():
    return
}
```

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,280 @@
...
+
+	content, runErr := d.runner.RunAction(reqCtx, req.Plugin, req.Action, req.Args)
+
+	fields := map[string]any{"content": content}
+	if runErr != nil {
+		fields["error"] = runErr.Error()
+	}
+
+	replyStream := fmt.Sprintf(replyStreamFmt, req.ID)
+	if err := d.client.XAdd(ctx, &redis.XAddArgs{
```
zhisme: **Blocker: reply lost on graceful shutdown.** XAdd uses parent `ctx`, but the `XAck` defer correctly uses a fresh context. Net effect during shutdown: message gets ack'd (plugin loses the retry) *and* the reply silently fails to deliver — plugin-side waiter hangs until its own timeout. Give the reply XAdd + the `Expire` below the same fresh-context treatment as the ack.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,280 @@
...
+			ChannelID: req.ChannelID,
+		}
+		reqCtx = profile.WithProfile(reqCtx, p)
+		reqCtx = actor.WithActor(reqCtx, req.EntityID)
+	}
+	if req.SessionID != "" {
+		reqCtx = actor.WithSessionID(reqCtx, req.SessionID)
+	}
+
+	content, runErr := d.runner.RunAction(reqCtx, req.Plugin, req.Action, req.Args)
```
zhisme: No per-request timeout on `RunAction`. A hung plugin action permanently occupies a semaphore slot; 32 hangs = dispatcher wedged with no visible failure. Wrap `reqCtx` with `context.WithTimeout` before calling the runner — make it operator-tunable alongside `dedup_ttl`.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,280 @@
...
+		client:     client,
+		runner:     runner,
+		consumerID: hostname,
+		sem:        make(chan struct{}, maxConcurrentHandlers),
+	}
+}
+
+// NewExecRedisClient creates a Redis client for the exec dispatcher.
+// Accepts the same cluster config values used for deduplication.
+func NewExecRedisClient(redisURL, masterName string, sentinels []string, password, sentinelPassword string) (redis.UniversalClient, error) {
```
zhisme: Opens a second Redis pool on top of the dedup client, both targeting the same Redis. Worth reusing one `UniversalClient` across both callers to halve connection count and simplify operator capacity planning.

docs/cluster.md
```diff
@@ -0,0 +1,109 @@
...
+cluster:
+  enabled: true
+  redis_url: "${REDIS_URL}"
+```
+
+## Modes
+
+| Mode | When to use | Required fields |
+|---|---|---|
+| Standalone | Single Redis instance or Redis Cluster | `redis_url` |
```
zhisme: **Blocker: misleading claim.** Code uses `redis.NewClient`, not `redis.NewClusterClient`. Against a real Redis Cluster, cross-slot `SET NX` will route by key hash but you lose cluster-aware pipelining/MOVED handling — this isn't what "Redis Cluster support" means. Either drop the claim or add a `NewClusterClient` path with its own constructor.

docs/cluster.md
```diff
@@ -0,0 +1,109 @@
+# Cluster Mode (Multi-Pod Deduplication)
+
+When running multiple OpenTalon pods (e.g. in Kubernetes with `replicas: 2+`), each pod connects independently to the channel plugin (e.g. Slack Socket Mode). Because Slack delivers every event to all open WebSocket connections from the same app, each pod receives every inbound message — which would result in duplicate responses.
+
+Setting `cluster.enabled: true` activates Redis-backed message deduplication. When a message arrives, each pod races to acquire a Redis lock (`SET NX EX`) keyed to the message's channel, conversation, and timestamp. Only the pod that wins the lock processes the message; the others silently skip it. If Redis is unreachable the lock attempt is logged as a warning and the message is processed anyway (fail-open), so a Redis outage never silences the bot.
```
zhisme, line 5: "Never silences the bot" understates the trade-off: if Redis is fully down, every one of N pods processes each message → N duplicate responses (for Slack, N messages in the channel). That's worse than silence for some operators. Name the trade-off explicitly and consider surfacing a dedup-error-rate metric so operators can alert on it.

cmd/opentalon/main.go
```diff
@@ -433,6 +462,32 @@ func main() {
...
+		if redisURL != "" || len(cfg.Cluster.Sentinels) > 0 {
+			execClient, err := plugin.NewExecRedisClient(
+				redisURL,
+				cfg.Cluster.MasterName,
+				cfg.Cluster.Sentinels,
+				cfg.Cluster.Password,
+				cfg.Cluster.SentinelPassword,
+			)
+			if err != nil {
+				slog.Warn("plugin exec dispatcher: redis connect failed, exec disabled", "error", err)
```
zhisme: Inconsistent failure policy: plugin-exec logs a warning and continues when Redis is unreachable, but dedup (line 429) hard-exits on the same failure. Pick one — operators seeing "cluster.enabled worked but plugin_exec silently disabled" will be confused. If warn-and-continue is the right call, make dedup consistent; otherwise both should fail fast.

internal/channel/registry.go
```diff
@@ -136,13 +155,36 @@ func (r *Registry) dispatch(ch pkg.Channel, inbox <-chan pkg.InboundMessage) {
...
 				ctx := logger.WithTraceID(r.ctx, traceID)
+
+				// Deduplication: when running multiple pods each pod receives every
+				// inbound message. Only the pod that wins the Redis SET NX lock
+				// processes the message; others silently skip it.
+				if dedup != nil {
+					if m.Timestamp.IsZero() {
+						slog.Warn("dedup skipped: message has no timestamp", "channel", ch.ID(), "session", sessionKey)
+					} else {
+						key := fmt.Sprintf("dedup:%s:%s:%d", m.ChannelID, m.ConversationID, m.Timestamp.UnixNano())
```
zhisme, line 177: Dedup key format is load-bearing — operators will grep Redis for it and `docs/cluster.md` documents it separately. Hoist to a named `const` or helper so it's defined in one place. Also: this uses `m.ChannelID` (wire value from the plugin), not `ch.ID()` (registered channel name). Probably fine since `ChannelID` is the provider-side ID, but worth a one-line comment naming the invariant so a future plugin author doesn't break it.

pkg/plugin/protocol.go
```diff
@@ -75,6 +85,9 @@ func ParseHandshake(line string) (Handshake, error) {
 	}
 	h.Network = parts[1]
 	h.Address = parts[2]
```
zhisme, line 88: `HTTPAddr` is accepted without validation. Plugins are trusted code so threat model is limited, but a typo like `:9091` or `bad:addr:format` registers a broken proxy that only surfaces at first request. A `net.SplitHostPort` check here costs nothing and fails fast at handshake time.

internal/dedup/redis_integration_test.go
```diff
@@ -0,0 +1,116 @@
+//go:build redis
```
zhisme, line 1: Standalone is well-covered; Sentinel (`NewSentinel`) has zero integration coverage. Sentinel is the production HA path — at minimum one test gated behind the `redis` tag against a Sentinel docker-compose would be valuable.

internal/plugin/exec_dispatcher_test.go
```diff
@@ -0,0 +1,152 @@
+package plugin
```
zhisme, line 1: Only `parseExecRequest` is unit-tested. The `handle` loop, semaphore, autoclaim, and context propagation (`profile.WithProfile` + `actor.WithActor`) are untested. Given this path can execute *any* plugin action impersonating any entity/group, it deserves at least one integration test against real Redis (reuse the `redis` build tag): end-to-end XADD → handle → XAck → reply-on-stream, plus a test asserting entity_id/group_id propagate into the reqCtx seen by the runner.

docs/workflows.md
```diff
@@ -0,0 +1,43 @@
...
+
+With a workflow plugin installed, users can create and manage multi-step automated workflows — for example, fetching Jira issues every morning and posting a summary to Slack — directly through chat or a REST API.
+
+## Plugin REST API (reverse proxy)
+
+Plugins can optionally expose their own HTTP API through OpenTalon's existing webhook server. Set `OPENTALON_HTTP_PORT` in the plugin's environment **and** add `expose_http: true` to the plugin's config. OpenTalon will then reverse-proxy `/{plugin-name}/*` to the plugin's HTTP server — no extra port or load-balancer config needed.
+
+The `expose_http: true` opt-in is required because the webhook server is typically internet-facing. Without explicit operator approval, a misconfigured plugin cannot accidentally publish its API.
+
+For example, with `opentalon-workflows` running on port 9091:
```
zhisme, line 29: Port `3978` appears here with no explanation of where it comes from — reader has to grep `webhookPortOrDefault` to learn it's the default webhook port. One-line note would help.

proto/plugin.proto
```diff
@@ -34,6 +34,9 @@ message PluginCapabilities {
   string name = 1;
   string description = 2;
   repeated Action actions = 3;
+  // Optional text injected into the LLM system prompt when this plugin is loaded.
```
zhisme, line 37: `system_prompt_addition` lets a plugin inject arbitrary text into the LLM system prompt. Fine given plugins are trusted subprocesses today, but worth naming explicitly in this doc comment: if plugin trust is ever relaxed (e.g. third-party plugin marketplaces), this becomes a prompt-injection vector.

zhisme (commented):
Bundling dedup + plugin-exec + HTTP proxy + system_prompt_addition into one PR makes this hard to review and bisect — consider splitting. Dedup path is tight; plugin-exec is a trust boundary that deserves its own scrutiny.

Blockers inline: shutdown hang on full semaphore, reply lost on graceful shutdown, misleading Redis Cluster doc claim.

zhisme (commented):
Overall shape is sound; test coverage is better than typical. Leaving inline comments on correctness, shutdown, and config-shape concerns. The shutdown + panic-recovery items are the ones I'd fix before merge — the rest are polish.

cmd/opentalon/main.go
```diff
@@ -433,6 +473,27 @@ func main() {
...
+				if d, err := time.ParseDuration(cfg.PluginExec.ActionTimeout); err == nil {
+					actionTimeout = d
+				} else {
+					slog.Warn("plugin_exec.action_timeout invalid, using default", "value", cfg.PluginExec.ActionTimeout)
+				}
+			}
+			dispatcher := plugin.NewExecDispatcher(sharedRedis, orch, actionTimeout)
+			dispatchCtx, dispatchCancel := context.WithCancel(ctx)
+			defer dispatchCancel()
+			go dispatcher.Start(dispatchCtx)
```
zhisme, line 497: Graceful shutdown is broken here. `Start` contains `defer wg.Wait()` to drain in-flight handlers, but nothing in main waits for `Start` to return. Sequence on SIGINT:

1. `dispatchCancel()` fires (deferred above)
2. `sharedRedis.Close()` fires (deferred earlier in main)
3. `Start` is still mid-drain — handlers blocked on `XAck` / `XAdd` against a closed client

Expose a `Wait()` (or make `Start` blocking, run it in a goroutine that signals a done channel, and wait on that after cancel) before `sharedRedis.Close()` runs.

cmd/opentalon/main.go
```diff
@@ -433,6 +473,27 @@ func main() {
 		slog.Info("channels loaded")
 	}
 
+	// Start plugin exec dispatcher (allows trusted plugins to execute ToolRegistry actions via Redis).
+	// Requires plugin_exec.enabled: true and cluster.redis_url (or sentinels) to be set.
+	if cfg.PluginExec.Enabled {
+		if sharedRedis == nil {
+			slog.Warn("plugin_exec.enabled is true but no redis_url configured, exec disabled")
```
zhisme: Cluster mode with `enabled: true` and no Redis config exits (good). Plugin-exec in the same situation only warns. Operators who set `plugin_exec.enabled: true` expecting workflows to work will get silent failure — their plugin will hang on the reply stream until its request timeout. Make this fatal at startup like cluster does.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,302 @@
...
+					defer wg.Done()
+					defer func() { <-d.sem }()
+					d.handle(ctx, m)
+				}(msg)
+			}
+		}
+	}
+}
+
+func (d *ExecDispatcher) handle(ctx context.Context, msg redis.XMessage) {
```
zhisme, line 206: Missing `defer recover()`. A panic in `RunAction` takes down the pod, not just the request. Every trusted-plugin bug becomes a cluster outage. Recover, log, write the panic back as an error on the reply stream, and let XAck in the existing deferred cleanup remove the message from the PEL — otherwise plugins hang until their own timeout.

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,302 @@
...
+// longer than autoclaimMinIdle (i.e. owned by a consumer that died).
+func (d *ExecDispatcher) autoclaimLoop(ctx context.Context, wg *sync.WaitGroup) {
+	ticker := time.NewTicker(autoclaimInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			msgs, _, err := d.client.XAutoClaim(ctx, &redis.XAutoClaimArgs{
```
zhisme: `XAutoClaim` returns `(msgs, nextCursor, err)` and the next cursor is thrown away. With >100 stale pending entries (default count), the tail is not reclaimed until the next 30s tick — and because `Start: "0-0"` always resets, the head of the PEL can repeatedly starve entries further in. Loop within a tick until `nextCursor == "0-0"` (or empty).

internal/plugin/exec_dispatcher.go
```diff
@@ -0,0 +1,302 @@
...
+		runner:        runner,
+		consumerID:    hostname,
+		sem:           make(chan struct{}, maxConcurrentHandlers),
+		actionTimeout: actionTimeout,
+	}
+}
+
+// NewExecRedisClient creates a Redis client for the exec dispatcher.
+// Accepts the same cluster config values used for deduplication.
+func NewExecRedisClient(redisURL, masterName string, sentinels []string, password, sentinelPassword string) (redis.UniversalClient, error) {
```
zhisme: Misnamed and in the wrong package. This is a generic Redis client factory — it's also used for dedup (see `sharedRedis` in main.go). The `dedup` package still has parallel `NewStandalone` / `NewSentinel` constructors that are now unreachable from main.go. Either collapse to one helper in `internal/redisclient` used by both subsystems, or delete the now-unused dedup constructors. Two code paths that ping Redis on startup is noise.

internal/channel/webhook_server.go
```diff
@@ -66,6 +69,36 @@ func (s *WebhookServer) register(port int, path string, handler http.HandlerFunc
...
+// All requests to /{prefix}/* are forwarded to http://{targetAddr}/* with the
+// prefix stripped. The server is started lazily on the first registration.
+func RegisterReverseProxy(port int, prefix, targetAddr string) error {
+	port = webhookPortOrDefault(port)
+	target, err := url.Parse("http://" + targetAddr)
+	if err != nil {
+		return fmt.Errorf("invalid proxy target %q: %w", targetAddr, err)
+	}
+	proxy := httputil.NewSingleHostReverseProxy(target)
+	pattern := "/" + strings.Trim(prefix, "/") + "/"
```
zhisme: Two things:

1. Pattern is `/{prefix}/`. A request to `/{prefix}` (no trailing slash) relies on Go's default 301 redirect from `http.ServeMux` — worth a test to confirm it works end-to-end, otherwise plugin users get a surprising 404.
2. Consider `http.StripPrefix` instead of the manual `TrimPrefix` + `RawPath` dance — stdlib handles the encoded-path case correctly and the code gets shorter.

internal/orchestrator/orchestrator.go
```diff
@@ -851,6 +851,10 @@ func (o *Orchestrator) buildSystemPrompt(ctx context.Context, userMessage string
 				fmt.Fprintf(&sb, "  - %s: %s%s\n", p.Name, p.Description, req)
 			}
 		}
+		if cap.SystemPromptAddition != "" {
+			sb.WriteString(cap.SystemPromptAddition)
```
zhisme: Plugin-supplied text concatenated raw into the system prompt has no visible boundary. A plugin can emit headings, fake delimiters, or instructions that blend with the surrounding prompt. Trusted today, but for attributability (and the third-party-plugin future called out in the proto comment) add a marker: e.g. `\n--- plugin: {cap.Name} ---\n` before the addition. Operators and the LLM can then see what came from where.

internal/plugin/manager.go
```diff
@@ -152,6 +154,18 @@ func (m *Manager) Load(ctx context.Context, entry PluginEntry) error {
 		m.watchProcess(ctx, entry.Name, proc)
 	}
 
+	// Reverse-proxy /{plugin-name}/* through the shared webhook server only when the
+	// operator explicitly opts in via expose_http: true. The plugin's declared HTTPAddr
+	// alone is not sufficient — the operator must have the final say since the webhook
+	// server is typically internet-facing.
+	if httpAddr := client.HTTPAddr(); httpAddr != "" && entry.ExposeHTTP {
```
zhisme: Four states from the `expose_http` × `OPENTALON_HTTP_PORT` matrix: (yes,yes) works, (yes,no) wastes resources (plugin runs an HTTP server nobody proxies), (no,yes) silently no-ops, (no,no) fine. States 2 and 3 are config mistakes with no signal. A warning when `client.HTTPAddr() != "" && !entry.ExposeHTTP` (and inverse) would catch these at startup.

pkg/plugin/protocol.go
```diff
@@ -75,6 +86,12 @@ func ParseHandshake(line string) (Handshake, error) {
...
+	if len(parts) == 4 {
+		h.HTTPAddr = parts[3]
+		if _, _, err := net.SplitHostPort(h.HTTPAddr); err != nil {
```
zhisme: `net.SplitHostPort` accepts `0.0.0.0:9091` and external hosts. `httpAddrFromEnv` in server.go hardcodes `127.0.0.1:` and the contract is loopback-only. Since plugins are trusted subprocesses this isn't a new trust boundary, but enforcing loopback-only here makes the contract explicit and prevents a plugin accidentally exposing itself outside the host.

internal/config/config.go
```diff
@@ -24,6 +24,32 @@ type Config struct {
 	Lua             *LuaConfig               `yaml:"lua,omitempty"`
 	Profiles        ProfilesConfig           `yaml:"profiles,omitempty"`
 	Bootstrap       BootstrapConfig          `yaml:"bootstrap,omitempty"`
+	Cluster         ClusterConfig            `yaml:"cluster,omitempty"`
+	PluginExec      PluginExecConfig         `yaml:"plugin_exec,omitempty"`
```
zhisme, line 29: Config shape is coupled: `plugin_exec.enabled` requires `cluster.redis_url` to be set. A user who wants workflows but not dedup still has to fill in a section named "cluster". Consider lifting Redis connection config to a top-level `redis:` block (both subsystems depend on it) or giving each section its own connection info with internal pool sharing.

PR #66, opened 2026-04-14
add channel output format

OpakAlex:


zhisme (commented):
Output-format plumbing looks solid. Main concerns are on the bundled `StripInternalBlocks` / retry logic — a few smells there and the scope feels like it wants to be a separate PR. Inline comments below.

internal/orchestrator/orchestrator.go
```diff
@@ -580,7 +581,23 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
 
 		calls := o.parser.Parse(resp.Content)
 		if calls == nil {
-			result.Response = resp.Content
+			stripped := StripInternalBlocks(resp.Content)
+			if stripped == "" && stripped != resp.Content {
```
zhisme: `stripped == "" && stripped != resp.Content` is inscrutable. The second clause is only false when `resp.Content == ""`, so write it plainly:

```go
if stripped == "" && resp.Content != "" {
```

Took me two passes to convince myself the original was correct.

internal/orchestrator/orchestrator.go
```diff
@@ -580,7 +581,23 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
...
+				_ = o.sessions.AddMessage(sessionID, provider.Message{
+					Role:    provider.RoleAssistant,
+					Content: resp.Content,
+				})
+				_ = o.sessions.AddMessage(sessionID, provider.Message{
+					Role:    provider.RoleUser,
+					Content: "[system] Your previous response contained tool call blocks that could not be executed. Please respond to the user in natural language without tool calls.",
+				})
+				log.Debug("LLM produced only unparseable tool calls, retrying for plain response", "round", i+1)
+				continue
```
zhisme, line 611: This retry shares the 20-iteration agent-loop budget. If the LLM persistently emits only unparseable tool-call blocks, it'll burn the whole budget before producing a reply. Cap strip-retries to 1–2 independently.

Also: you persist the bad assistant message + a synthetic `[system] ...` user message into the session. Those artifacts become part of the conversation history and will leak into future summaries/replays. Either mark them with metadata so they can be filtered, or keep them transient instead of `AddMessage`.

internal/orchestrator/orchestrator.go
```diff
@@ -868,9 +885,40 @@ func (o *Orchestrator) buildSystemPrompt(ctx context.Context, userMessage string
...
+// provides a custom ResponseFormatPrompt it takes precedence over the built-in
+// hints. Returns empty string when no format is configured.
+func channelFormatHint(ctx context.Context) string {
+	caps := pkgchannel.CapabilitiesFromContext(ctx)
+	if caps.ResponseFormatPrompt != "" {
+		return caps.ResponseFormatPrompt
+	}
+	switch caps.ResponseFormat {
+	case pkgchannel.FormatSlack:
+		return "You are responding in a Slack channel. Format your reply using Slack mrkdwn: *bold*, _italic_, `inline code`, ```code blocks```, >blockquote, and bullet lists with •. Do not use standard Markdown syntax (no ** or ##)."
```
zhisme: "bullet lists with •" — Slack renders `-`/`*` bullets natively; forcing `•` is likely worse than the default. Drop that clause or switch to `-`.

Also worth noting in docs that format hints are best-effort; LLMs routinely ignore them.

internal/orchestrator/parser.go
```diff
@@ -197,6 +197,51 @@ func stripSurroundingQuotes(s string) string {
...
+			// No closing tag — drop everything after the opening tag.
+			break
+		}
+	}
+	return sb.String()
+}
+
+// StripToolCallBlocks is an alias kept for compatibility with existing callers.
+// Prefer StripInternalBlocks which also strips [plugin_output] blocks.
+func StripToolCallBlocks(s string) string { return StripInternalBlocks(s) }
```
zhisme: `StripToolCallBlocks` is "kept for compatibility with existing callers" — but there are no existing callers (grep confirms). Delete it. Don't plant dead compatibility shims for code that never shipped.

internal/orchestrator/parser.go
```diff
@@ -197,6 +197,51 @@ func stripSurroundingQuotes(s string) string {
...
+// StripInternalBlocks removes any internal protocol blocks from s
+// ([tool_call]...[/tool_call] and [plugin_output]...[/plugin_output]).
+// Used to clean up LLM responses so raw protocol content is never forwarded
+// to the channel (e.g. when the LLM echoes plugin output or uses an
+// unparseable tool name).
+func StripInternalBlocks(s string) string {
+	for _, tags := range internalBlockTags {
+		s = stripTaggedBlocks(s, tags[0], tags[1])
+	}
+	return strings.TrimSpace(s)
```
zhisme, line 218: `TrimSpace` on the whole string is a subtle behavior change vs. the prior `result.Response = resp.Content`. Probably fine, but worth noting — e.g. a reply like `"  plain answer  "` now has surrounding whitespace stripped.

pkg/channel/types.go
```diff
@@ -36,15 +36,45 @@ type FileAttachment struct {
...
+	Reactions            bool           `yaml:"reactions" json:"reactions"`
+	Edits                bool           `yaml:"edits" json:"edits"`
+	MaxMessageLength     int64          `yaml:"max_message_length" json:"max_message_length"`
+	ResponseFormat       ResponseFormat `yaml:"response_format" json:"response_format"`
+	ResponseFormatPrompt string         `yaml:"response_format_prompt" json:"response_format_prompt"`
+}
+
+type contextKey int
+
+const capabilitiesKey contextKey = iota
```
zhisme: `type contextKey int` with a single `iota` value is overkill. Idiomatic Go uses `struct{}` (or a named empty-struct type) so there's no confusion about numbering:

```go
type capabilitiesKey struct{}
```

Minor.

internal/channel/yaml_spec.go
```diff
@@ -26,11 +27,13 @@ type YAMLChannelSpec struct {
...
-	Files            bool `yaml:"files"`
-	Reactions        bool `yaml:"reactions"`
-	Edits            bool `yaml:"edits"`
-	MaxMessageLength int  `yaml:"max_message_length"`
+	Threads              bool               `yaml:"threads"`
+	Files                bool               `yaml:"files"`
+	Reactions            bool               `yaml:"reactions"`
+	Edits                bool               `yaml:"edits"`
+	MaxMessageLength     int                `yaml:"max_message_length"`
+	ResponseFormat       pkg.ResponseFormat `yaml:"response_format"`
```
zhisme, line 35: No validation of `response_format` at load time. An operator typo like `response_format: SLACK` or `response_format: slak` silently falls into the `default` case in `channelFormatHint` (no hint at all). Validate against the known set in the YAML parser and return an error for unknown values — operators will otherwise file bugs about hints not working.

internal/channel/yaml_spec.go
```diff
@@ -26,11 +27,13 @@ type YAMLChannelSpec struct {
...
+	ResponseFormatPrompt string             `yaml:"response_format_prompt"`
```
zhisme, line 36: `response_format_prompt` is injected verbatim into the system prompt with no length cap. YAML is trusted config so it's not a security issue, but a runaway prompt from a copy-paste will silently bloat every LLM request. A sanity length check + log warning is cheap.

internal/orchestrator/orchestrator.go
```diff
@@ -580,7 +585,27 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
...
+			stripped := StripInternalBlocks(resp.Content)
+			if stripped == "" && resp.Content != "" {
+				// The LLM produced only unparseable tool call blocks — nothing to send
+				// to the user. Ask it to retry in plain language, but cap at one such
+				// retry to avoid burning the whole agent-loop budget. The correction
+				// messages are kept transient (not persisted to the session) so they
+				// don't pollute future summaries or replays.
+				if stripRetries >= 1 {
+					log.Debug("LLM repeatedly produced only unparseable tool calls, giving up on strip-retry", "round", i+1)
+					result.Response = resp.Content
```
zhisme: Bug: when we give up after the retry, this sends `resp.Content` raw — which by construction here contains *only* unparseable `[tool_call]...[/tool_call]` blocks (that's exactly the branch we're in). So the user sees the raw protocol garbage that this whole code path was added to suppress.

Either send `stripped` (empty string — let the caller treat it as no-reply) or substitute a generic fallback like `"(no response)"`. Don't pass through the broken content.

internal/orchestrator/orchestrator.go
```diff
@@ -580,7 +585,27 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
...
+					return result, nil
```
zhisme, line 603: Early `return result, nil` here also skips `o.sessions.AddMessage(...)` and `o.maybeRecordWorkflow(...)` that the normal `calls == nil` path runs a few lines below. Session history and workflow recording become inconsistent on the give-up path. Probably want to fall through to the same tail rather than returning early.

zhisme (commented):
Round 2. All 8 prior comments addressed cleanly — retry cap + transient messages, `struct{}` context key, format validation with error + tests, length warning, dead alias removed, Slack `-` bullets, condition simplified. Teams/WhatsApp/Discord additions are scope creep but small, tested, and consistent.

One real bug in the give-up branch and a small consistency nit. Nothing else.

PR #67, opened 2026-04-14
support files formats for anthropic

OpakAlex:


internal/provider/anthropic.go
```diff
@@ -216,18 +216,32 @@ func (p *AnthropicProvider) toAnthMessage(m Message) (anthMessage, error) {
 
 	var blocks []anthContentBlock
 	for _, f := range m.Files {
```
zhisme, line 219: The doc comment on `toAnthMessage` (just above, line 207) is now stale — it still says "all others map to document blocks", but this switch has three branches. Please update it to describe image / pdf / text.

internal/provider/anthropic.go
```diff
@@ -216,18 +216,32 @@ func (p *AnthropicProvider) toAnthMessage(m Message) (anthMessage, error) {
...
+			})
+		case f.MimeType == "application/pdf":
+			blocks = append(blocks, anthContentBlock{
+				Type: "document",
+				Source: &anthImageSource{
+					Type:      "base64",
+					MediaType: f.MimeType,
+					Data:      base64.StdEncoding.EncodeToString(f.Data),
+				},
+			})
```
zhisme, line 238: This `default` branch will silently corrupt any binary, non-PDF attachment (`.docx`, `.zip`, `.xlsx`, audio, etc.). `string(f.Data)` on non-UTF-8 bytes gets garbled by JSON marshaling (invalid bytes become U+FFFD), and the model receives noise with no error.

Before this PR, Anthropic would reject such uploads with an API error — now it looks like it worked. That's a regression in signal quality.

Options:
- Return an error for unsupported MIME types (safest).
- Allowlist text-like MIMEs (`text/*`, `application/json`, `application/xml`) and error on the rest.
- At minimum, guard with `utf8.Valid(f.Data)` and error otherwise.

Separately: Anthropic supports `text/plain` as a `document` block with a `text`-typed source (not base64). That preserves the "this is an attachment" framing instead of inlining the file as if the user typed it — worth considering as the representation for text files.

internal/provider/anthropic_test.go
```diff
@@ -208,3 +208,82 @@ func TestAnthropicDefaultBaseURL(t *testing.T) {
...
+			}
+			if !tc.wantSource && blocks[0].Source != nil {
+				t.Error("expected no source block, got one")
+			}
+			if tc.wantText != "" && blocks[0].Text != tc.wantText {
+				t.Errorf("text = %q, want %q", blocks[0].Text, tc.wantText)
+			}
+			if tc.wantSource && blocks[0].Source != nil && blocks[0].Source.MediaType != tc.mimeType {
+				t.Errorf("media_type = %q, want %q", blocks[0].Source.MediaType, tc.mimeType)
+			}
```
zhisme, line 320: Missing unhappy-path coverage — exactly where this PR introduces risk:

- Binary non-PDF (e.g. `application/zip` with non-UTF-8 bytes) — what should happen? Currently it silently mangles, and no test pins down the intended behavior.
- Multiple mixed-type files in one message.
- Empty `Content` with a single file (does the text block get appended?).

The happy-path table is fine as far as it goes, but without the unhappy path these tests give false confidence on the default branch.

zhisme (commented):
File-format dispatch is a good direction, but the default-to-text fallback silently corrupts binary uploads. See inline comments.

internal/provider/anthropic.go
```diff
@@ -216,18 +218,38 @@ func (p *AnthropicProvider) toAnthMessage(m Message) (anthMessage, error) {
...
+		case strings.HasPrefix(f.MimeType, "text/"),
+			f.MimeType == "application/json",
+			f.MimeType == "application/xml":
+			blocks = append(blocks, anthContentBlock{
+				Type: "document",
+				Source: &anthImageSource{
+					Type: "text",
+					Data: string(f.Data),
```
zhisme: Worth verifying against the Anthropic API: for plain-text document sources, their spec requires `media_type: "text/plain"` on the source, e.g.

```
"source": { "type": "text", "media_type": "text/plain", "data": "..." }
```

This code omits `media_type` entirely (struct has `omitempty` and nothing is set). If the API rejects sources without `media_type`, every text/csv/json/xml upload will 400. Please send a real request through and confirm — if it's required, hardcode `"text/plain"` here since Anthropic's text source only understands that one media type regardless of the original MIME.

Also: tests assert `wantMediaType == ""` for text sources, which pins the *current* behavior but wouldn't catch this if it turns out to be wrong. Once you verify, update the test to assert the correct value.

internal/provider/anthropic_test.go
```diff
@@ -208,3 +208,176 @@ func TestAnthropicDefaultBaseURL(t *testing.T) {
...
+				t.Fatal(err)
+			}
+
+			var blocks []anthContentBlock
+			if err := json.Unmarshal(msg.Content, &blocks); err != nil {
+				t.Fatal(err)
+			}
+			// blocks[0] is the file block, blocks[1] is the text message
+			if len(blocks) != 2 {
+				t.Fatalf("blocks = %d, want 2", len(blocks))
```
zhisme, line 286: Nit: `application/xml` is in the code path but not covered by the table. One more row would close that gap.

zhisme (commented):
Round 2: all three points from last review addressed well — error on unknown MIME, text types now use `document` block with `text` source, unhappy-path tests added (unsupported MIME, mixed files, empty content). One concern inline about the text source payload, otherwise LGTM.

internal/provider/file.go
```diff
@@ -0,0 +1,38 @@
...
+	FileClassPDF
+	// FileClassBinary covers everything else. Providers that cannot handle
+	// arbitrary binary data should return an error for this class.
+	FileClassBinary
+)
+
+// ClassifyFile determines the FileClass for a file attachment from its MIME type.
+// Any type not explicitly recognised is treated as FileClassBinary; the caller
+// is responsible for deciding whether to surface an error to the user.
+func ClassifyFile(mimeType string, _ []byte) FileClass {
```
zhisme, line 25: Nit: the `data []byte` parameter is unused (`_ []byte`) and every caller has to pass `f.Data` anyway. Either use it (e.g. `utf8.Valid(data)` to downgrade mislabeled `text/*` that's actually binary) or drop the parameter. Unused params designed for "future use" tend to stay unused — easier to add later when there's a real caller.

internal/provider/openai.go
```diff
@@ -168,17 +171,20 @@ func (p *OpenAIProvider) Stream(_ context.Context, _ *CompletionRequest) (Respon
 	return nil, fmt.Errorf("streaming not yet implemented for provider %s", p.id)
 }
 
-func (p *OpenAIProvider) toOAIRequest(req *CompletionRequest) oaiRequest {
+func (p *OpenAIProvider) toOAIRequest(req *CompletionRequest) (oaiRequest, error) {
 	msgs := make([]oaiMessage, len(req.Messages))
 	for i, m := range req.Messages {
+		if len(m.Files) > 0 {
+			return oaiRequest{}, fmt.Errorf("provider %s does not support file attachments", p.id)
```
zhisme, line 178: Nit on the error message: OpenAI's API does support image attachments via vision — the limitation is in this provider's implementation, not the provider itself. Something like `"file attachments not yet implemented for provider %s"` reads more accurately. Not worth blocking.

zhisme (approved):
Final review — approving. Core behavior is correct:

- Text source now sets `media_type: "text/plain"` with a comment explaining why (the API rejects anything else). Tests updated to assert it.
- `ClassifyFile` helper cleanly centralizes the MIME dispatch and is well covered by `file_test.go`.
- OpenAI provider now surfaces an error instead of silently dropping attachments — nice catch while you were here.

Two nits, non-blocking:

PR #68, opened 2026-04-14
prometheus support

OpakAlex:


cmd/opentalon/main.go
```diff
@@ -14,13 +14,16 @@ import (
 
 	"github.com/redis/go-redis/v9"
 
+	"net/http"
```
zhisme: `net/http` is stdlib — belongs in the stdlib import group above `redis/go-redis`, not in its own group. `goimports` will sort.

cmd/opentalon/main.go
```diff
@@ -76,6 +79,24 @@ func main() {
...
+		metricsCollector = metrics.New()
+		addr := cfg.Metrics.Addr
+		if addr == "" {
+			addr = ":9090"
+		}
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", metricsCollector.Handler())
+		go func() {
+			slog.Info("metrics server listening", "addr", addr)
+			if err := http.ListenAndServe(addr, mux); err != nil && err != http.ErrServerClosed {
```
zhisme: `http.ListenAndServe` with the default server has no timeouts (gosec G114). Even for an internal scrape endpoint, prefer:

```go
srv := &http.Server{Addr: addr, Handler: mux, ReadHeaderTimeout: 5 * time.Second}
if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed { ... }
```

Also no graceful shutdown path — goroutine outlives `main`'s context. Fine for now, but inconsistent with the rest of the lifecycle.

cmd/opentalon/main.go
```diff
@@ -76,6 +79,24 @@ func main() {
 	}
 	logger.Setup(logLevel)
 
+	// Start Prometheus metrics server if enabled.
+	var metricsCollector *metrics.Collector
+	if cfg.Metrics.Enabled {
+		metricsCollector = metrics.New()
+		addr := cfg.Metrics.Addr
+		if addr == "" {
+			addr = ":9090"
```
zhisme: Default `:9090` is applied here, but the doc comment on `MetricsConfig.Addr` says it lives in config. Move the fallback into `Parse` alongside the other defaults for one source of truth.

Separately: `:9090` collides with Prometheus's own default — your own docker-compose has to remap to `9091:9090` to work around it. Consider `:2112` (common Go app-metrics convention).

cmd/opentalon/main.go
```diff
@@ -564,19 +592,25 @@ func (a *usageRecorderAdapter) RecordUsage(ctx context.Context, entityID, groupI
...
-		SessionID:    sessionID,
-		ModelID:      modelID,
-		InputTokens:  inputTokens,
-		OutputTokens: outputTokens,
-		ToolCalls:    toolCalls,
-		InputCost:    inputCostUSD,
-		OutputCost:   outputCostUSD,
-	}); err != nil {
-		slog.Warn("usage record failed", "entity", entityID, "error", err)
+	if a.store != nil {
```
zhisme, line 598: Brittle — `RecordUsage` now assumes the caller guarded `store != nil || collector != nil` in `main.go`. Add an early-exit so the adapter is self-contained:

```go
if a.store == nil && a.collector == nil {
    return
}
```

internal/metrics/collector.go
```diff
@@ -0,0 +1,109 @@
...
+			Help: "Total LLM input tokens consumed.",
+		}, []string{"model", "channel", "group"}),
+
+		llmOutputTokens: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "opentalon_llm_output_tokens_total",
+			Help: "Total LLM output tokens produced.",
+		}, []string{"model", "channel", "group"}),
+
+		llmCostUSD: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "opentalon_llm_cost_usd_total",
```
zhisme: Prom style typically prefers separate metrics over a label that changes the value's semantic. Splitting into `opentalon_llm_input_cost_usd_total` / `opentalon_llm_output_cost_usd_total` would mirror how input/output tokens are already split. Not a blocker.

internal/metrics/collector.go
```diff
@@ -0,0 +1,109 @@
...
+			Help: "Total LLM spend in USD.",
+		}, []string{"model", "channel", "group", "type"}),
+
+		llmRequests: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "opentalon_llm_requests_total",
```
zhisme: `llm_requests_total` is ambiguous — a single orchestrator run can make multiple calls to the LLM (tool-loop turns). Doc says "completed orchestrator runs"; consider `opentalon_orchestrator_runs_total` to avoid confusion.

internal/metrics/collector.go
```diff
@@ -0,0 +1,109 @@
...
+	labels := prometheus.Labels{
+		"model":   modelID,
+		"channel": channelID,
+		"group":   groupID,
+	}
+	c.llmInputTokens.With(labels).Add(float64(inputTokens))
+	c.llmOutputTokens.With(labels).Add(float64(outputTokens))
+	c.llmRequests.With(labels).Inc()
+
+	if inputCostUSD > 0 {
```
zhisme: `inputCostUSD > 0` skips the series when cost is zero. Free-tier models therefore never appear in `opentalon_llm_cost_usd_total`, so `sum()` queries silently drop them. Either drop the guard (cost=0 series is fine) or document the behavior in `docs/prometheus-metrics.md`.

internal/orchestrator/orchestrator_test.go
```diff
@@ -1561,3 +1581,163 @@ func TestBuildSystemPromptIncludesFormatSection(t *testing.T) {
...
+		Name:        "jira",
+		Description: "Jira integration",
+		Actions:     []Action{{Name: "create_issue", Description: "Create issue"}},
+	}, &echoExecutor{})
+
+	memory := state.NewMemoryStore("")
+	sessions := state.NewSessionStore("")
+	sessions.Create("test-session-obs")
+
+	opts.UsageRecorder = nil
```
zhisme: `opts.UsageRecorder = nil` silently overwrites what the caller passed in. Footgun for future tests — either drop this line or rename the helper to signal it zeroes the field.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,178 @@
...
+
+Add a `metrics` section to your `config.yaml`:
+
+```yaml
+metrics:
+  enabled: true
+  addr: ":9090"   # optional; defaults to :9090
+```
+
+When enabled, OpenTalon starts an HTTP server on the configured address. Prometheus can then scrape it at `http://<host>:9090/metrics`.
```
zhisme: Add a line: `/metrics` is unauthenticated — bind to loopback or a private interface if the network isn't trusted.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,178 @@
...
+    - port: metrics
+      interval: 30s
+      path: /metrics
+```
+
+## Example PromQL queries
+
+```promql
+# Total input tokens per model (rate over 5m)
+rate(opentalon_llm_input_tokens_total[5m])
```
zhisme: `rate(opentalon_llm_input_tokens_total[5m])` returns a per-`(model,channel,group)` rate, not "total input tokens per model". Either fix the comment or wrap with `sum by (model) (...)`.

zhisme (commented):
Inline comments below. Overall: clean implementation with good test coverage. Nits around server timeouts, import grouping, docs, and PromQL examples.

zhisme (approved):
All 10 comments from the previous round addressed. Verified:

- ✅ `net/http` moved into stdlib import group
- ✅ `ReadHeaderTimeout: 5s` + graceful `Shutdown` on signal
- ✅ Default `:2112` moved into `Parse`, with `TestParseMetricsEnabledDefaultAddr` covering it
- ✅ `usageRecorderAdapter.RecordUsage` early-exits when both sinks are nil
- ✅ Cost metric split into `opentalon_llm_input_cost_usd_total` / `opentalon_llm_output_cost_usd_total`
- ✅ `llm_requests_total` renamed to `orchestrator_runs_total`
- ✅ Zero-cost guard dropped; series always emitted (test renamed to `TestRecordUsageZeroCostEmitsSeries`)
- ✅ `opts.UsageRecorder = nil` removed from test helper
- ✅ Docs: security warning added, port references updated to `:2112`, PromQL rate wrapped in `sum by (model)`, USD total query updated for split metrics

One tiny nit (non-blocking): `metricsSrv.Shutdown(context.Background())` in `cmd/opentalon/main.go` uses an un-cancellable context — if an in-flight scrape hangs, so does shutdown. A 5s `context.WithTimeout` would be safer but it's not worth another round. LGTM.

PR #72, opened 2026-04-16
implement browser login

OpakAlex:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/opentalon/opentalon
 
-go 1.24.0
+go 1.25
```
zhisme, line 3: Unrelated to the plugin protocol change — same for the 5 hits in `.github/workflows/ci.yml` and the 2 in `Dockerfile`. Please split the Go bump into its own PR so either change can be reverted independently if CI breaks on 1.25.

pkg/plugin/convert.go
```diff
@@ -19,9 +19,11 @@ func capsToProto(c CapabilitiesMsg) *pluginpb.PluginCapabilities {
 			}
 		}
 		actions[i] = &pluginpb.Action{
-			Name:        a.Name,
-			Description: a.Description,
-			Parameters:  params,
+			Name:              a.Name,
+			Description:       a.Description,
+			Parameters:        params,
+			UserOnly:          a.UserOnly,
```
zhisme, line 25: Worth calling out in the commit message: before this PR, `UserOnly` was not propagated from `ActionMsg` into the proto, so JSON-mode plugins couldn't declare `user_only` actions at all. This silently fixes that. Real behavior change — don't bury it under a commit titled "implement browser login".

zhisme (commented):
Plumbs `inject_context_args` through both plugin protocols (proto + JSON) and quietly fixes a `UserOnly` propagation bug in `capsToProto`. Also bumps Go 1.22/1.24 → 1.25.

**Title/scope mismatch.** "implement browser login" — but there's no login code here, only protocol scaffolding. Rewrite the title/commit to describe what actually landed; `git log` shouldn't lie.

**Fail-open injection concern** (pre-existing, not introduced here, but newly relevant given the stated use case). In `internal/orchestrator/orchestrator.go:1088-1093` the injection loop only overwrites `call.Args[name]` when a provider is registered *and* returns a non-empty value:

```go
for _, name := range action.InjectContextArgs {
    if provide := o.contextArgProviders[name]; provide != nil {
        if v := provide(ctx, name); v != "" {
            args[name] = v
        }
    }
}
```

If a plugin declares `inject_context_args: ["actor_id"]` and (a) no provider is registered or (b) the provider returns `""`, an LLM-supplied `actor_id` in `call.Args` passes through unchanged. For identity args feeding auth flows — which is presumably the browser-login use case — that's spoofable by the LLM.

Options: fail closed (reject the call if a declared arg has no provider), or strip the key from `args` before the provider loop so missing ⇒ absent rather than LLM-supplied. Worth a follow-up before this plumbing is wired to anything auth-sensitive.

**Tests** are clean — both mapping directions, present/absent cases. No end-to-end check that a declaring plugin actually receives the injected value at Execute time, but each half is independently covered.

PR #77, opened 2026-04-20
implement llm limits

OpakAlex:


internal/profile/verifier.go
```diff
@@ -173,8 +186,10 @@ func (v *Verifier) cleanupLoop() {
 }
 
 // Verify returns the Profile for token, using the cache when valid.
+// channelType is the originating channel ID (e.g. "slack", "telegram") and is
+// forwarded to the WhoAmI server if ChannelTypeHeader is configured.
 // Returns ErrAuthFailed if the server rejects the token or returns incomplete data.
-func (v *Verifier) Verify(ctx context.Context, token string) (*Profile, error) {
+func (v *Verifier) Verify(ctx context.Context, token, channelType string) (*Profile, error) {
 	key := sha256.Sum256([]byte(token))
```
zhisme: Cache key is `sha256(token)` only — it ignores `channelType`. If the WhoAmI server returns channel-specific data (which is the whole point of `ChannelTypeHeader`), the first caller's channel wins for `CacheTTL`. A user hitting Slack first then web will see Slack's `limit`/`model` on web until the entry expires.

Fix: include `channelType` in the key, e.g. `sha256(token + "\x00" + channelType)`. As-is, `ChannelTypeHeader` is latently broken for anyone who wants to use it.

zhisme (commented):
Line-by-line comments below. Ship-blockers in my view: the cache-key bug (verifier.go) and the silent limit parse (verifier.go). The rest is fine to fix in follow-ups.

internal/profile/verifier.go
```diff
@@ -281,13 +299,28 @@ func (v *Verifier) callServer(ctx context.Context, token string) (*Profile, erro
 	}
 
 	model := jsonString(raw[v.cfg.ModelField])
+	channelTypeResp := jsonString(raw[v.cfg.ChannelTypeField])
+
+	var limit int
+	if lraw, ok := raw[v.cfg.LimitField]; ok {
+		_ = json.Unmarshal(lraw, &limit)
```
zhisme: Silent unmarshal: if the server sends `"limit": "1000"` (string) the field becomes 0 → unlimited. Fail-open on a *limit* parse error is the wrong default — malformed auth-server output shouldn't silently remove enforcement.

At minimum log a warning; arguably return `ErrAuthFailed` since the profile data is malformed. Same applies to `time.ParseDuration(s)` below.

internal/profile/verifier.go
```diff
@@ -281,13 +299,28 @@ func (v *Verifier) callServer(ctx context.Context, token string) (*Profile, erro
...
+	}
+	var limitWindow time.Duration
+	if ltraw, ok := raw[v.cfg.LimitTimeField]; ok {
+		if s := jsonString(ltraw); s != "" {
+			limitWindow, _ = time.ParseDuration(s)
```
zhisme: Same concern as `limit` above: `"limit_time": "1hour"` (invalid) silently becomes 0 → unlimited. Log it at least.

internal/profile/types.go
```diff
@@ -4,12 +4,17 @@
...
-	Plugins   []string // plugin IDs returned by WhoAmI (auto-saved to group_plugins table)
-	ChannelID string   // set by the handler; used for usage statistics
-	Model     string   // optional model override returned by WhoAmI (e.g. "anthropic/claude-3-5-sonnet-20241022")
+	EntityID    string        // stable identifier from the WhoAmI server (used for session/memory scoping)
+	Group       string        // group name from the WhoAmI server (used for tool-access control)
+	Token       string        // original bearer token (may be forwarded to MCP servers)
+	Plugins     []string      // plugin IDs returned by WhoAmI (auto-saved to group_plugins table)
+	ChannelID   string        // set by the handler; used for usage statistics
+	Model       string        // optional model override returned by WhoAmI (e.g. "anthropic/claude-3-5-sonnet-20241022")
+	ChannelType string        // optional channel type returned by WhoAmI (e.g. "slack", "web", "api")
```
zhisme, line 17: `ChannelType` is populated from the response but nothing reads it in this PR. Either wire it up (and point at where) or drop it.

Also: naming is confusing next to `ChannelID`. The handler sets `ChannelID = msg.ChannelID`, `ChannelType` comes from the WhoAmI response, and the handler passes `msg.ChannelID` to `Verify` as the `channelType` argument. Three names for overlapping concepts.

internal/channel/handler.go
```diff
@@ -33,13 +40,26 @@ func NewMessageHandler(
...
 				slog.Warn("profile verification failed", "error", err, "channel", msg.ChannelID)
 				return errorResponse(msg, "authentication failed"), nil
 			}
 			p.ChannelID = msg.ChannelID
 			ctx = profile.WithProfile(ctx, p)
+
+			// Enforce per-profile token spend limit when configured.
+			if limitChecker != nil && p.Limit > 0 && p.LimitWindow > 0 {
+				since := time.Now().Add(-p.LimitWindow)
+				used, lerr := limitChecker.TotalTokensSince(ctx, p.EntityID, since)
```
zhisme: Race: two concurrent requests from the same entity can both read `used < limit` and both proceed. For a token-spend limit this is probably fine (small overshoot), but flagging — don't extend this pattern to anything hard-capped like billing without a reserve/commit step.

internal/channel/handler.go
```diff
@@ -33,13 +40,26 @@ func NewMessageHandler(
...
+				if lerr != nil {
+					slog.Warn("limit check failed", "error", lerr, "entity", p.EntityID)
+				} else if used >= p.Limit {
+					slog.Info("token limit exceeded", "entity", p.EntityID, "used", used, "limit", p.Limit, "window", p.LimitWindow)
+					return errorResponse(msg, "Token limit reached. Please try again later."), nil
```
zhisme: Style inconsistency: `"Token limit reached. Please try again later."` is capitalized with trailing period. Other handler errors here are lowercase no-period (`"profile token required"`, `"authentication failed"`). Pick one convention.

internal/state/store/usage.go
```diff
@@ -32,6 +32,22 @@ func NewUsageStore(db *DB) *UsageStore {
 	return &UsageStore{db: db}
 }
 
+// TotalTokensSince returns the sum of input + output tokens for entityID
+// recorded on or after since. Used to enforce per-profile token limits.
+func (s *UsageStore) TotalTokensSince(ctx context.Context, entityID string, since time.Time) (int, error) {
+	sinceStr := since.UTC().Format(time.RFC3339)
+	row := s.db.SQLDB().QueryRowContext(ctx, s.db.Dialect().Rebind(`
+		SELECT COALESCE(SUM(input_tokens + output_tokens), 0)
```
zhisme, line 40: `SUM(...)` runs on every message. With the new index it's O(rows-in-window-for-entity), fine now. At scale (millions of rows/entity/window) you'd want a rolling counter or bucketed aggregation. Not a blocker — just noting for later.

cmd/opentalon/main.go
```diff
@@ -465,7 +469,7 @@ func main() {
 		}
 	}
 	runner := &channelRunner{orch: orch}
-	handler := channel.NewMessageHandler(ensureSession, runner, orch.RunAction, toolRegistry.HasAction, profileVerifier)
+	handler := channel.NewMessageHandler(ensureSession, runner, orch.RunAction, toolRegistry.HasAction, profileVerifier, usageStore)
```
zhisme: Six positional args is getting hard to read. Consider a struct arg on the next addition — passing `verifier, nil` vs `nil, checker` is easy to misread.

PR #79, opened 2026-04-20
allow dynamic creds

OpakAlex:


proto/plugin.pb.go
```diff
@@ -0,0 +1,536 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
zhisme: Delete this file. It's a duplicate of `proto/pluginpb/plugin.pb.go` generated at the wrong output path. Nothing imports `github.com/opentalon/opentalon/proto` — `internal/plugin/client.go:9` uses `.../proto/pluginpb`. 536 lines of dead code. Same for `proto/plugin_grpc.pb.go`. Probably caused by running `protoc` without the usual `go_opt=paths=source_relative` (or whatever the project's gen script uses) — worth re-running the canonical codegen command.

proto/plugin_grpc.pb.go
```diff
@@ -0,0 +1,200 @@
+// Code generated by protoc-gen-go-grpc. DO NOT EDIT.
```
zhisme: Same as `proto/plugin.pb.go` — delete. Dead duplicate of `proto/pluginpb/plugin_grpc.pb.go`.

internal/plugin/client.go
```diff
@@ -104,11 +105,16 @@ func (c *Client) Execute(ctx context.Context, call orchestrator.ToolCall) orches
 
 // ExecuteContext is like Execute but respects context cancellation.
 func (c *Client) ExecuteContext(ctx context.Context, call orchestrator.ToolCall) orchestrator.ToolResult {
+	var credentials map[string]string
+	if p := profile.FromContext(ctx); p != nil && len(p.Credentials) > 0 {
+		credentials = p.Credentials
+	}
```
zhisme, line 111: Trust-boundary concern: this forwards the *entire* credentials map to every plugin. If the user has `{timly: ..., jira: ...}` and the `jira` plugin is invoked, it still receives the `timly` token. One buggy/compromised plugin sees every other service's token.

Suggest scoping:

```go
if p := profile.FromContext(ctx); p != nil {
    if tok, ok := p.Credentials[call.Plugin]; ok {
        credentials = map[string]string{call.Plugin: tok}
    }
}
```

If the fan-out is intentional (e.g. a plugin legitimately needs multiple service tokens), add a comment saying so — and mirror the security note you added on `SystemPromptAddition`.

Minor: `len(p.Credentials) > 0` is redundant — a nil map marshals identically to an empty one over proto.

internal/profile/verifier.go
```diff
@@ -333,6 +337,11 @@ func (v *Verifier) callServer(ctx context.Context, token, channelType string) (*
 		}
 	}
 
+	var credentials map[string]string
```
zhisme, line 342: Silent `_ = json.Unmarshal(...)` swallows malformed JSON. If the WhoAmI server returns the wrong shape (e.g. an array, or `{"timly": 123}`), the user gets nil credentials with no signal of why. Consistent with other fields in this file, but still a debug hole — one line of `log.Warn` (or whatever the project uses) would save someone an hour later.

Not blocking.

internal/profile/verifier_test.go
```diff
@@ -378,6 +378,88 @@ func TestVerifier_ChannelTypeResponse(t *testing.T) {
 	}
 }
```
zhisme, line 380: Coverage gap: no test for a malformed `credentials` field (array instead of object, non-string values, etc.). Given the unmarshal error is swallowed, a test pinning the behavior — "malformed credentials yields nil, no error" — would at least document the contract.

zhisme (changes requested):
Two things to fix before merge:

1. **Duplicate generated proto at the wrong path** — `proto/plugin.pb.go` and `proto/plugin_grpc.pb.go` are 736 lines of dead code; the real files live in `proto/pluginpb/`. Nothing imports `.../proto`. Looks like an accidental `protoc` invocation with the wrong output dir.
2. **Credential fan-out** — every plugin invocation receives the full credentials map. If the `jira` plugin is called, it also gets the `timly` token. Scope to `credentials[call.Plugin]` unless this is intentional.

Nits below. Tests on the happy paths are solid — missing coverage is malformed-JSON handling and cross-plugin scoping.

PR #80, opened 2026-04-20
dynamic plugins

OpakAlex:


internal/orchestrator/orchestrator.go
```diff
@@ -1135,16 +1135,15 @@ func (o *Orchestrator) executeCall(ctx context.Context, call ToolCall) ToolResul
...
-			slog.Warn("BLOCKED tool call for restricted plugin", "plugin", call.Plugin, "action", call.Action)
-			return ToolResult{
-				CallID: call.ID,
-				Error:  fmt.Sprintf("plugin %q is not available for this profile", call.Plugin),
-			}
+	// Defense-in-depth: block plugins that are not allowed for this profile.
+	// In strict mode (WhoAmI plugins list) every plugin is checked; in non-strict
+	// mode only capabilities with AllowedGroups set are gated (pluginAllowed handles both).
+	// This mirrors buildSystemPrompt and protects against crafted tool calls.
+	if allowed := o.resolveAllowedPlugins(ctx); !o.pluginAllowed(capForCheck, allowed) {
```
zhisme: Strict mode probably blocks guards/preparers/pipelines. `executeCall` is the shared dispatch for LLM calls **and** internal plumbing: `runSinglePreparer` (line 361), `runGuardPlugins` via line 698, and `executePipeline` (line 726) all land here. For a user whose WhoAmI returns `Plugins: ["mymcp"]`, a configured XSS guard or content preparer gets rejected with `"plugin %q is not available for this profile"` — and there's no graceful fallback for guards, so the request fails.

Same goes for `permission_plugin`: line 1108 skips the *permission check* for it, but the allowlist check here still fires. If it's not in the WhoAmI list, it gets blocked.

Two fixes worth considering:
1. Exempt non-`FromLLM` calls from the allowlist (matches how `UserOnly` on line 1150 and `rejectUnknownArgs` on line 1165 already treat internal calls as trusted).
2. Union configured guard/preparer/permission plugin names into the allowed set.

Option 1 is cleaner. Either way, add a strict-mode test that exercises a guard — right now `plugin_filter_test.go` only covers LLM-originated calls.

internal/orchestrator/orchestrator.go
```diff
@@ -1338,62 +1337,79 @@ type allowedPluginsKey struct{}
...
+		return cachedAllowedPlugins{m: map[string]bool{}}
 	}
 	ids, err := o.groupPluginLookup.PluginsForGroup(ctx, p.Group)
 	if err != nil {
 		slog.Warn("group plugin lookup failed", "group", p.Group, "error", err)
-		return nil
-	}
-	if len(ids) == 0 {
-		return map[string]bool{} // group exists but has no plugins — return empty (all restricted plugins blocked)
+		return cachedAllowedPlugins{}
```
zhisme, line 1403: DB failure returns `cachedAllowedPlugins{}` → `pluginAllowed` treats it as unrestricted. Fails open. Same behavior as before (was `return nil`), but strict mode widens the blast radius — a WhoAmI outage now silently grants access to every plugin instead of just the `AllowedGroups`-gated ones. Probably still the right choice (closed-fail would brick the bot during incidents), but worth a comment here saying so explicitly.

internal/orchestrator/plugin_filter_test.go
```diff
@@ -0,0 +1,294 @@
...
+	for _, msg := range llm.requests[1].Messages {
+		if strings.Contains(msg.Content, "not available") {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Error("second LLM request should contain the 'not available' block message")
+	}
+}
```
zhisme, line 347: Missing: a strict-mode test where a configured guard/preparer is not in `Profile.Plugins`. That's where the regression from the executeCall comment above would surface. End-to-end `Run` with an `OrchestratorOpts.Guards` entry targeting `public` and `Profile.Plugins: ["mymcp"]` — expect the guard to run, not to return `"not available for this profile"`.

pkg/plugin/grpc_server_test.go
```diff
@@ -0,0 +1,132 @@
...
+	if token, ok := creds["mymcp"]; ok {
+		merged["Authorization"] = "Bearer " + token
+	}
+
+	if merged["X-Tenant-ID"] != "acme-corp" {
+		t.Errorf("X-Tenant-ID = %q, want acme-corp", merged["X-Tenant-ID"])
+	}
+	if merged["Authorization"] != "Bearer per-user-token" {
+		t.Errorf("Authorization = %q, want 'Bearer per-user-token'", merged["Authorization"])
+	}
```
zhisme: This test constructs the `merged` map inside the test itself and then asserts on it — no production code path performs that merge, so this is a doc-style test rather than a behavior test. If the merge is the intended pattern for plugin implementers, move it into a helper in the plugin package and test the helper; otherwise consider dropping this and keeping `TestGRPCServer_CredentialsReachHandler` as the real coverage.

internal/orchestrator/orchestrator.go
```diff
@@ -1338,62 +1337,79 @@ type allowedPluginsKey struct{}
...
+// resolveAllowedPlugins returns the set of plugin IDs allowed for the current profile.
+// Returns a zero cachedAllowedPlugins (m == nil, strict == false) when no profile is set
+// or no group plugin lookup is configured (= no restrictions).
 // The result is cached in ctx by Run; within a single Run call this never hits the DB twice.
-func (o *Orchestrator) resolveAllowedPlugins(ctx context.Context) map[string]bool {
+//
+// Priority:
+//  1. Profile.Plugins non-nil → strict allowlist from WhoAmI (per-request, every plugin gated)
+//  2. Group DB lookup → non-strict (only capabilities with AllowedGroups are gated)
+func (o *Orchestrator) resolveAllowedPlugins(ctx context.Context) cachedAllowedPlugins {
```
zhisme, line 1370: Minor: the zero `cachedAllowedPlugins` value now carries two meanings — "cache miss" (caller path distinguishes via the type-assertion `ok`) and "no restrictions" (m == nil). It works, but a reader has to cross-check the type-assertion at line 1361 to see they aren't conflated. Not blocking; a comment or a sentinel would make it self-evident.

zhisme (commented):
Overall: the credential wire-through is a real bug fix — `internal/plugin/client.go:117` was already sending them and `requestFromProto` was dropping them. The strict-mode change is where I'd want agreement before merging; see inline comments.

PR #81, opened 2026-04-20
STT support

OpakAlex:


internal/orchestrator/orchestrator.go
```diff
@@ -307,6 +310,79 @@ func (o *Orchestrator) handlePreparerFailure(prep ContentPreparerEntry, details
...
+	if !hasSTT || len(audioFiles) == 0 {
+		return content, files
+	}
+
+	for _, prep := range o.preparers {
+		if !prep.STT {
+			continue
+		}
+		for _, af := range audioFiles {
+			transcript, err := o.runSTTPreparer(ctx, prep, af)
```
zhisme, line 437: **Bug: multi-STT-preparer transcribes every audio file N times.**

Outer loop iterates all STT preparers, inner loop iterates all audio files — so 2 preparers × 1 file → 2 transcripts prepended. Audio files should be consumed (removed from `audioFiles`) once a preparer handles them, or only the first STT preparer should run.

```go
// Option: break after first STT preparer that succeeds for each file
// or convert audioFiles to a map and delete after success
```

internal/orchestrator/orchestrator.go
```diff
@@ -307,6 +310,79 @@ func (o *Orchestrator) handlePreparerFailure(prep ContentPreparerEntry, details
...
+	}
+	result := o.executeCall(ctx, call)
+	if result.Error != "" {
+		return "", fmt.Errorf("stt plugin %q: %s", prep.Plugin, result.Error)
+	}
+	return result.Content, nil
+}
+
 func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPreparerEntry, content, callPrefix string, allowInvoke bool) (string, *RunResult, error) {
 	if strings.HasPrefix(prep.Plugin, "lua:") {
```
zhisme, line 492: **Bug: `remaining` gets duplicate audio entries with multiple fail-open preparers.**

`remaining = append(remaining, af)` runs once per failing STT preparer. Two preparers both failing with `fail_open=true` → same audio file appended twice → LLM receives the file twice.

zhisme, line 492: **Bug: mixed state when first preparer succeeds, second fail-opens.**

Scenario: 2 STT preparers, 1 audio file. Preparer 1 succeeds → transcript in content, audio not in `remaining`. Preparer 2 fails with `fail_open=true` → audio re-appended to `remaining`. LLM sees both the transcript text AND the raw audio file.

internal/orchestrator/orchestrator.go
```diff
@@ -307,6 +310,79 @@ func (o *Orchestrator) handlePreparerFailure(prep ContentPreparerEntry, details
...
+		ID:     fmt.Sprintf("stt-%s-%s", prep.Plugin, prep.Action),
+		Plugin: prep.Plugin,
+		Action: prep.Action,
+		Args: map[string]string{
+			"file_data": base64.StdEncoding.EncodeToString(f.Data),
+			"file_mime": f.MimeType,
+		},
+	}
+	result := o.executeCall(ctx, call)
+	if result.Error != "" {
```
zhisme, line 483: No size guard before encoding. A large audio file (e.g. 50 MB) doubles peak memory here (raw bytes copied to base64 string). Add a max file size check at the ingestion boundary before `MessageFile` is populated.

internal/orchestrator/stt_test.go
```diff
@@ -0,0 +1,263 @@
+package orchestrator
```
zhisme, line 1: Missing tests for:
- multiple STT preparers (would expose the bugs flagged above)
- multiple audio files in one request
- `runSTTPreparer` with unknown plugin/action (`HasAction` returning false)

zhisme (changes requested):
Overall structure is clean — the preparer extension point works well for STT. The multi-preparer loop has correctness bugs that need fixing before merge. Single-preparer path (the common case) works correctly and is well tested.

PR #96, opened 2026-04-22
Add response_formatters post-processing plugin system (#96)

OpakAlex:
Add a response_formatters hook symmetric with content_preparers that runs after the LLM response. Supports gRPC plugins and Lua scripts. Ships a built-in Lua formatter (scripts/format-response.lua) that converts standard Markdown to Slack mrkdwn, Teams, WhatsApp, HTML, Telegram, and plain text — fixing output from cheap models that ignore format hints.

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -144,7 +144,7 @@ type Orchestrator struct {
 	guard                   *Guard
 	rules                   *RulesConfig
 	preparers               []ContentPreparerEntry
-	formatters              []ResponseFormatterEntry       // run after final response; text-in/text-out
+	formatters              []ResponseFormatterEntry      // run after final response; text-in/text-out
 	guards                  []ContentPreparerEntry        // subset of preparers with Guard:true; run before every LLM call
 	luaScriptPaths          map[string]string             // optional; plugin name -> path to .lua script (for "lua:name" preparers)
 	permissionChecker       PermissionChecker             // optional; when set, executeCall checks permission before running
@@ -419,7 +419,7 @@ func (o *Orchestrator) formatResponse(ctx context.Context, result *RunResult) er
 			scriptName := strings.TrimPrefix(f.Plugin, "lua:")
 			scriptPath := o.luaScriptPaths[scriptName]
 			if scriptPath == "" {
-				err = fmt.Errorf("Lua script path not found for %q", scriptName)
+				err = fmt.Errorf("lua script path not found for %q", scriptName)
 			} else {
 				formatted, err = lua.RunFormat(scriptPath, result.Response, responseFormat)
 			}
@@ -456,6 +456,9 @@ func (o *Orchestrator) formatResponse(ctx context.Context, result *RunResult) er
 			}
 			return fmt.Errorf("response formatter %s: %w", name, err)
 		}
+		// Guard: ignore empty results so a broken formatter can't blank the
+		// response. A formatter that intentionally returns "" is treated as a
+		// no-op, preserving the previous response text.
 		if formatted != "" {
 			result.Response = formatted
 		}
@@ -533,7 +536,7 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
 			res, err := o.executePipeline(ctx, sessionID, p)
 			if err == nil && res != nil {
 				if fmtErr := o.formatResponse(ctx, res); fmtErr != nil {
-					return nil, fmt.Errorf("response formatter: %w", fmtErr)
+					return nil, fmtErr
 				}
 			}
 			return res, err
@@ -723,7 +726,7 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
 			})
 			o.maybeRecordWorkflow(ctx, result, userMessage)
 			if fmtErr := o.formatResponse(ctx, result); fmtErr != nil {
-				return nil, fmt.Errorf("response formatter: %w", fmtErr)
+				return nil, fmtErr
 			}
 			return result, nil
 		}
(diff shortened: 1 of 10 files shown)


scripts/format-response.lua
```diff
@@ -0,0 +1,238 @@
...
+      table.insert(parts, transform_fn(seg.text))
+    end
+  end
+  return table.concat(parts)
+end
+
+-- Slack mrkdwn: **bold** -> *bold*, ## Heading -> *Heading*
+local function format_slack(s)
+  -- Headings: # ... or ## ... -> *...*
+  s = s:gsub("(\n?)#+%s+([^\n]+)", function(nl, heading)
```
zhisme: **Bug: heading regex doesn't anchor to line start.** `(\n?)#+%s+` — the `\n?` is optional, so any `#` with a space after it in mid-text matches. Examples:

- `"run # as a delimiter"` → `"run *as a delimiter*"`
- `"look at # 5 below"` → `"look at *5 below*"`

Start-of-string is already handled separately by the `s:sub(1,1) == "#"` block below, so this branch should require a real newline: drop the `?` (`(\n)#+%s+...`).

Same bug in `format_teams`, `format_whatsapp`, `format_html` (all use the same pattern). Worth a table-driven test covering `("run # foo", "slack") → "run # foo"`.

scripts/format-response.lua
```diff
@@ -0,0 +1,238 @@
...
+  end)
+  -- Handle heading at start of string (no preceding newline)
+  if s:sub(1, 1) == "#" then
+    s = s:gsub("^#+%s+([^\n]+)", function(heading)
+      heading = heading:gsub("%*%*(.-)%*%*", "%1")
+      return "*" .. heading .. "*"
+    end)
+  end
+  -- Bold: **text** -> *text* (non-greedy)
+  s = s:gsub("%*%*(.-)%*%*", "*%1*")
```
zhisme: Slack italic is lossy. Markdown `*italic*` stays as `*italic*` after this, which Slack mrkdwn renders as bold (Slack italic is `_text_`). Input `**bold** *italic*` becomes `*bold* *italic*` — both render bold.

Either convert `*x*` → `_x_` before the `**x**` → `*x*` step, or document the limitation. Not a regression (current behavior), but a surprise given the script's purpose is fixing Markdown for Slack.

internal/orchestrator/orchestrator.go
```diff
@@ -390,6 +401,68 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
+				call := ToolCall{
+					ID:     fmt.Sprintf("formatter-%s-%s", f.Plugin, f.Action),
+					Plugin: f.Plugin,
+					Action: f.Action,
+					Args: map[string]string{
+						"text":            result.Response,
+						"response_format": responseFormat,
+					},
+				}
+				toolResult := o.executeCall(ctx, call)
```
zhisme, line 439: `executeCall` runs `permissionChecker` if configured. Formatters aren't LLM-driven tools — if a deployment has a permission plugin that restricts tool access, formatter actions will go through the same check and may be silently denied (returned as `toolResult.Error`, then treated as formatter failure).

`runSinglePreparer` bypasses permission checks for preparers for this reason. Consider the same for formatters, or document that the permission plugin must whitelist formatter actions.

internal/orchestrator/orchestrator.go
```diff
@@ -390,6 +401,68 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
+			if f.Action != "" {
+				name += "." + f.Action
+			}
+			if f.FailOpen {
+				slog.Warn("response formatter failed, skipping", "formatter", name, "error", err)
+				continue
+			}
+			return fmt.Errorf("response formatter %s: %w", name, err)
+		}
+		if formatted != "" {
```
zhisme, line 462: `if formatted != ""` silently drops an intentionally-empty result. Probably desired as a guard against broken formatters returning empty, but worth a one-line comment saying that's deliberate — otherwise a reader will wonder why a formatter that wants to blank the response can't.

internal/orchestrator/orchestrator.go
```diff
@@ -390,6 +401,68 @@ func (o *Orchestrator) runSinglePreparer(ctx context.Context, prep ContentPrepar
...
+		if err != nil {
+			name := f.Plugin
+			if f.Action != "" {
+				name += "." + f.Action
+			}
+			if f.FailOpen {
+				slog.Warn("response formatter failed, skipping", "formatter", name, "error", err)
+				continue
+			}
+			return fmt.Errorf("response formatter %s: %w", name, err)
```
zhisme, line 457: Double-wrap: this returns `"response formatter my-plugin.format: <err>"`, then the caller (lines 536, 726) wraps again as `"response formatter: %w"`. Final message: `response formatter: response formatter my-plugin.format: ...`. Drop one of the prefixes.

internal/orchestrator/orchestrator_test.go
```diff
@@ -1975,3 +1975,190 @@ func TestPluginCallObserverCalledForEachCallInMultiStep(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if result.Response != "untouched" {
+		t.Errorf("Response = %q, want %q", result.Response, "untouched")
+	}
+}
+
+func TestResponseFormatterLuaScript(t *testing.T) {
+	// Test Lua formatter via the orchestrator (requires a temp script file)
+	import_os := false
```
zhisme: Dead code — `import_os := false; _ = import_os` looks like a leftover. Also this test's name is misleading: it only tests the missing-script error path, not Lua end-to-end through the orchestrator. Rename to `TestResponseFormatterLuaMissingScript`.

zhisme (commented):
Left inline comments. Top-priority items: heading regex doesn't anchor to start-of-line (mid-text `#` gets converted), and `executeCall` subjects formatters to the permission plugin if one is configured — preparers bypass this for a reason.

PR #99, opened 2026-04-22
feat(metrics): expose entity_id label on LLM/orchestrator metrics

mscdit:
## Summary

The orchestrator already passes `entityID` to `UsageRecorder.RecordUsage`
(see `internal/orchestrator/orchestrator.go:626`), but the Prometheus
collector currently discards it. This means LLM spend and usage cannot
be attributed to individual actors (e.g. the Slack user ID that sent
the message).

This PR adds `entity_id` as a fourth label on the five LLM/orchestrator
counter vectors:

- `opentalon_llm_input_tokens_total`
- `opentalon_llm_output_tokens_total`
- `opentalon_llm_input_cost_usd_total`
- `opentalon_llm_output_cost_usd_total`
- `opentalon_orchestrator_runs_total`

`sessionID` remains discarded — session IDs rotate frequently and would
blow up Prometheus cardinality; session-level attribution is better
served by tracing.

## Compatibility

The new label is **appended** at the end of the label set, so existing
Prometheus queries that `sum()` across labels or filter by
`{model, channel, group}` keep working unchanged. Dashboards need no
updates unless operators want to add the new dimension.

## Cardinality note

For deployments with a bounded user base (internal Slack workspaces,
enterprise tenants) the label is safe. Operators running OpenTalon
against public channels with unbounded users can drop the label via
Prometheus `metric_relabel_configs`. This is now documented in
`docs/prometheus-metrics.md`.

## Test plan

- [x] `go test ./internal/metrics/...` — all green
- [x] `go build ./...` — all green
- [x] `go test ./...` — full suite green

### DIFF
--- a/docs/prometheus-metrics.md
+++ b/docs/prometheus-metrics.md
@@ -20,11 +20,11 @@ When enabled, OpenTalon starts an HTTP server on the configured address. Prometh
 
 | Metric | Type | Labels | Description |
 |--------|------|--------|-------------|
-| `opentalon_llm_input_tokens_total` | Counter | `model`, `channel`, `group` | Total input (prompt) tokens sent to the LLM |
-| `opentalon_llm_output_tokens_total` | Counter | `model`, `channel`, `group` | Total output (completion) tokens received from the LLM |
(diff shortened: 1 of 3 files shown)


zhisme (approved):
👍

Issue #97, opened 2026-04-22
Knowledge-Augmented RAG System


OpakAlex:
# Knowledge-Augmented RAG System for OpenTalon

## Context

Currently, ALL registered MCP tool definitions are dumped into the system prompt for every LLM call, regardless of relevance. For cheap/small models this wastes tokens and provides no domain context about *how* the product works. The goal:

1. On startup, auto-sync all plugin capabilities (Jira, GitLab, etc.) into Weaviate so they're searchable
2. User message triggers vector search → fetches relevant knowledge articles + relevant MCP actions
3. Only relevant tools + knowledge context sent to the LLM (not everything)
4. LLM can ask follow-up questions to the knowledge base mid-conversation without bothering the user
5. External systems can POST knowledge articles via an HTTP endpoint

**Example**: User says "assign all items in location A to person Bla" → system searches Weaviate → finds articles about how assignment works + relevant MCP actions (query locations, items, persons, create assignments) → LLM gets focused context → can query knowledge base for more details if needed.

---

## Architecture

```
Startup:
  pluginManager.LoadAll() → registry has all capabilities
       ↓
  weaviate.sync_actions(capabilities JSON) → MCPActions collection populated

External system / admin:
  POST http://opentalon:3978/weaviate/articles → KnowledgeArticles collection
  (also: knowledge_dir scanning on startup)

Runtime:
  User message
      ↓
  [prepare action] ── hybrid search ──→ Weaviate
      │                                    ├── KnowledgeArticles
      │                                    └── MCPActions
      ↓
  Enriched message + relevant_tools list
      ↓
  [buildSystemPrompt] ── filters tools to only relevant ones
      ↓
  LLM (focused context)
      ↓
  LLM can call weaviate.ask_knowledge for more info
```

---

### Files to modify
- `/private/var/www/opentalon/weaviate-plugin/handler.go`
- `/private/var/www/opentalon/weaviate-plugin/handler_test.go`
- `/private/var/www/opentalon/weaviate-plugin/main.go`

### 1a. New Config (no backward compat needed)

```go
type Config struct {
    Host         string            `json:"host"`
    Scheme       string            `json:"scheme"`
    Collections  CollectionsConfig `json:"collections"`  // required
    Fields       []string          `json:"fields"`
    Limit        int               `json:"limit"`
    KnowledgeDir string            `json:"knowledge_dir"` // optional: dir of .md files to auto-ingest
}

type CollectionsConfig struct {
[...]

PR #100, opened 2026-04-23
feat(orchestrator): add knowledge-augmented RAG core support (#97)

OpakAlex:
Add orchestrator-side support for the weaviate-plugin's RAG system:

- Preparer responses can now return `relevant_tools` to filter the system prompt to only the tools relevant to the user's query, reducing token usage on LLM calls.
- Built-in `allowed_plugins` context arg provider so RAG preparers and LLM-called actions (e.g. ask_knowledge) receive the profile's plugin allowlist for access-controlled filtering.
- SyncActions() syncs all registered plugin capabilities to the vector store at startup for retrieval-based tool filtering.
- IngestKnowledgeDir() optionally scans a directory of .md files and ingests them at startup; articles can also arrive via HTTP API.
- New `orchestrator.knowledge` config section for YAML-based setup.

### DIFF
--- a/cmd/opentalon/main.go
+++ b/cmd/opentalon/main.go
@@ -454,6 +454,12 @@ func main() {
 
 	// Sync plugin capabilities to the vector store and ingest knowledge articles
 	// from the configured directory. Both are fire-and-forget at startup.
+	//
+	// These calls use guard.ExecuteWithTimeout directly (not executeCall) because
+	// there is no actor/session at startup. This intentionally skips permission
+	// checks, audit logging, arg validation, and plugin-allowed filtering.
+	// If the sync or ingest plugin ever declares AuditLog=true, those calls
+	// will not be logged — acceptable for host-initiated startup work.
 	go func() {
 		orch.SyncActions(ctx)
 		orch.IngestKnowledgeDir(ctx)
--- a/cmd/opentalon/main.go
+++ b/cmd/opentalon/main.go
@@ -443,7 +443,21 @@ func main() {
 		GroupPluginLookup:       groupPluginStore,
 		UsageRecorder:           usageRecorder,
 		PluginCallObserver:      pluginObserver,
+		SyncActionsPlugin:       cfg.Orchestrator.Knowledge.SyncPlugin,
+		SyncActionsAction:       cfg.Orchestrator.Knowledge.SyncAction,
+		Knowledge: orchestrator.KnowledgeConfig{
+			Plugin: cfg.Orchestrator.Knowledge.Plugin,
+			Action: cfg.Orchestrator.Knowledge.Action,
+			Dir:    cfg.Orchestrator.Knowledge.Dir,
+		},
 	})
+
+	// Sync plugin capabilities to the vector store and ingest knowledge articles
+	// from the configured directory. Both are fire-and-forget at startup.
+	go func() {
+		orch.SyncActions(ctx)
+		orch.IngestKnowledgeDir(ctx)
+	}()
 
 	// Scheduler: wired after orchestrator so it can route job actions through orch.
 	// Personal reminders bypass the approver policy via AddPersonalJob.
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -87,6 +87,13 @@ type PluginCallObserver interface {
 	ObservePluginCall(plugin, action string, failed bool, inputTokens, outputTokens int)
 }
 
+// KnowledgeConfig configures the knowledge directory scanning feature.
+type KnowledgeConfig struct {
+	Plugin string // plugin name to call for ingestion (e.g. "weaviate")
+	Action string // action name for single-article ingestion (e.g. "ingest")
+	Dir    string // directory to scan for .md files
+}
+
 // OrchestratorOpts holds optional configuration for NewWithRules. Zero values mean defaults (no permission check, no summarization).
 type OrchestratorOpts struct {
 	CustomRules             []string
@@ -108,6 +115,9 @@ type OrchestratorOpts struct {
 	GroupPluginLookup       GroupPluginLookup  // optional; when set, filters tool list by profile group
 	UsageRecorder           UsageRecorder      // optional; when set, records LLM usage after each run
 	PluginCallObserver      PluginCallObserver // optional; when set, notified after each plugin/tool call
+	SyncActionsPlugin       string             // optional; plugin name for action sync (e.g. "weaviate")
+	SyncActionsAction       string             // optional; action name for sync (e.g. "sync_actions"); requires SyncActionsPlugin
+	Knowledge               KnowledgeConfig    // optional; knowledge directory ingestion
 }
 
 // MemoryStoreInterface is the scoped memory store used for general + per-actor memories.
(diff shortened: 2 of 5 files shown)


cmd/opentalon/main.go
```diff
@@ -443,8 +443,22 @@ func main() {
...
+		Knowledge: orchestrator.KnowledgeConfig{
+			Plugin: cfg.Orchestrator.Knowledge.Plugin,
+			Action: cfg.Orchestrator.Knowledge.Action,
+			Dir:    cfg.Orchestrator.Knowledge.Dir,
+		},
 	})
 
+	// Sync plugin capabilities to the vector store and ingest knowledge articles
+	// from the configured directory. Both are fire-and-forget at startup.
+	go func() {
```
zhisme, line 463: Fire-and-forget bypasses `executeCall`. `SyncActions`/`IngestKnowledgeDir` hit `guard.ExecuteWithTimeout` directly, skipping validate/sanitize/permission/audit. Startup has no actor so it's fine today, but if the sync or ingest plugin ever declares `AuditLog=true` the call won't be logged. Add a short comment here noting the intentional bypass.

internal/orchestrator/orchestrator.go
```diff
@@ -549,15 +591,24 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
...
 		if err != nil {
 			return nil, err
 		}
 		if blocked != nil {
 			return blocked, nil
 		}
 		content = guardedContent
+		// Last preparer that returns relevant_tools wins.
+		if len(tools) > 0 {
+			relevantTools = tools
```
zhisme, line 607: Last preparer wins for `relevant_tools`. Fine as a rule, but surprising when two preparers both compute relevance (the second silently overrides). Worth a one-liner in the config example so operators know ordering matters.

internal/orchestrator/orchestrator.go
```diff
@@ -998,11 +1058,25 @@ func (o *Orchestrator) buildSystemPrompt(ctx context.Context, userMessage string
...
 			if preparerAction[cap.Name+"."+action.Name] || action.UserOnly {
 				continue
 			}
+			if len(relevantToolSet) > 0 && !relevantToolSet[cap.Name+"."+action.Name] {
+				continue
+			}
+			visibleActions = append(visibleActions, action)
+		}
+		// Skip plugin header entirely if no actions are visible after filtering.
+		if len(visibleActions) == 0 && cap.SystemPromptAddition == "" {
```
zhisme, line 1075: No test covers this header-skip path. If a preparer returns `relevant_tools` strings that don't match any registered action, every plugin filters to zero visible actions and the header is skipped — worth one test to pin the behavior.

internal/orchestrator/orchestrator.go
```diff
@@ -1658,6 +1747,146 @@ func (p *permissionCheckerImpl) Allowed(ctx context.Context, actorID, plugin str
...
+	}
+	if !o.registry.HasAction(o.syncActionsPlugin, o.syncActionsAction) {
+		slog.Warn("sync_actions target not found", "plugin", o.syncActionsPlugin, "action", o.syncActionsAction)
+		return
+	}
+
+	caps := o.registry.ListCapabilities()
+	for _, cap := range caps {
+		// Don't sync the sync plugin itself.
+		if cap.Name == o.syncActionsPlugin {
```
zhisme, line 1780: This skips the entire sync plugin, not just the `sync_actions` action. If weaviate exposes other RAG-adjacent actions (e.g. the `ingest` wired below), they won't be discoverable via RAG either. If intentional, update the comment from "Don't sync the sync plugin itself" to "Don't sync any action from the sync plugin". If not intentional, compare against `o.syncActionsAction` per-action instead.

internal/orchestrator/orchestrator.go
```diff
@@ -1658,6 +1747,146 @@ func (p *permissionCheckerImpl) Allowed(ctx context.Context, actorID, plugin str
...
+			slog.Warn("sync_actions failed", "plugin", cap.Name, "error", result.Error)
+		} else {
+			slog.Info("sync_actions done", "plugin", cap.Name, "actions", len(entries))
+		}
+	}
+}
+
+// IngestKnowledgeDir scans the configured knowledge directory for .md files and
+// ingests each one via the configured plugin action. The directory is optional —
+// knowledge articles can also be submitted via the plugin's HTTP API. Idempotent;
```
zhisme: "Idempotent" claim depends entirely on the plugin. Weaviate upserting by title = idempotent; append-only = duplicates on re-run. Either drop the word or qualify it ("idempotent when the ingest action upserts by title").

internal/orchestrator/orchestrator.go
```diff
@@ -1658,6 +1747,146 @@ func (p *permissionCheckerImpl) Allowed(ctx context.Context, actorID, plugin str
...
+	// Dir is optional — articles may arrive exclusively via HTTP API.
+	if k.Dir == "" {
+		return
+	}
+	if !o.registry.HasAction(k.Plugin, k.Action) {
+		slog.Warn("knowledge ingest target not found", "plugin", k.Plugin, "action", k.Action)
+		return
+	}
+
+	entries, err := os.ReadDir(k.Dir)
```
zhisme: `os.ReadDir` is non-recursive — `knowledge/faq/billing.md` is silently skipped. Either document the top-level-only contract in the config example, or switch to `filepath.WalkDir`.

internal/orchestrator/orchestrator.go
```diff
@@ -1658,6 +1747,146 @@ func (p *permissionCheckerImpl) Allowed(ctx context.Context, actorID, plugin str
...
+		slog.Warn("knowledge ingest executor not found", "plugin", k.Plugin)
+		return
+	}
+
+	var count int
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasSuffix(e.Name(), ".md") {
+			continue
+		}
+		path := k.Dir + "/" + e.Name()
```
zhisme: Use `filepath.Join(k.Dir, e.Name())`. Trailing-slash safe and cross-platform.

internal/orchestrator/orchestrator.go
```diff
@@ -1658,6 +1747,146 @@ func (p *permissionCheckerImpl) Allowed(ctx context.Context, actorID, plugin str
...
+		data, err := os.ReadFile(path)
```
zhisme, line 1867: No size cap. `os.ReadFile` on an arbitrary `.md` in a user-configured dir → one multi-GB file = OOM at startup. Add a sanity limit (e.g. 10MB) and log-skip oversized files.

internal/orchestrator/orchestrator.go
```diff
@@ -1658,6 +1747,146 @@ func (p *permissionCheckerImpl) Allowed(ctx context.Context, actorID, plugin str
...
+		}
+	}
+	if count > 0 {
+		slog.Info("knowledge dir ingested", "dir", k.Dir, "files", count)
+	}
+}
+
+// mustGetExec returns the executor for a plugin or panics. Only used for
+// plugins whose existence was verified before the call.
+func mustGetExec(reg *ToolRegistry, name string) PluginExecutor {
```
zhisme: `mustGetExec` panic is unnecessary — both call sites already verified existence via `HasAction`/`GetExecutor`. Panicking for a "can't happen" path adds a failure mode without buying safety. Return `(nil, false)` and log a warn, matching the rest of the file's style.

zhisme (commented):
Review: knowledge-augmented RAG core. Main concerns inline. LGTM after `filepath.Join`, drop `mustGetExec` panic, file-size cap, and comment fixes — rest are judgment calls.

Issue #105, opened 2026-04-27
Subprocess Support for Core Orchestrator


OpakAlex:
Context

 The orchestrator currently runs a single linear agent loop: LLM → parse tool calls → execute sequentially → loop. There's no way for the LLM to "fork" a focused sub-task (vector search, ask another LLM,
 multi-step research) and get results back without polluting its main conversation context.

 Inspired by Unix process model and Claude Code's sub-agents: the main process (PID 1) can fork() child processes that inherit capabilities but have their own scope, run independently, and return results to the
 parent. This keeps the main conversation focused while enabling parallel-capable sub-tasks.

 Architecture: Unix Kernel Analogy

 Main Process (PID 0)          ← the user's conversation
   ├── fork() → Child (PID 1)  ← "search knowledge base for refund policy"
   ├── fork() → Child (PID 2)  ← "check what APIs are available for billing"
   └── waits for children, collects results, continues reasoning

 - Parent process = main agent loop in Run()
 - Child process = mini agent loop with ephemeral session, scoped tools, own iteration budget
 - fork() = LLM calls _subprocess.run tool
 - exit() = child produces final answer (no more tool calls)
 - waitpid() = parent receives child result as ToolResult
 - Process limits = max depth (no fork bombs), max iterations, timeout

 Implementation Plan

 1. New file: internal/orchestrator/subprocess.go

 Contains all subprocess logic in one file:

 Types:
 type SubprocessConfig struct {
     Enabled        bool          // master switch (default false)
     MaxDepth       int           // max nesting depth (default 2, hard cap 3)
     MaxIterations  int           // default iterations per child (default 5, hard cap 10)
     DefaultTimeout time.Duration // per-subprocess timeout (default 60s)
 }

 type subprocessRequest struct {
     Task          string   // focused task description
     AllowedTools  []string // optional plugin.action allowlist
     MaxIterations int      // 0 = use default
 }

 Context depth tracking via context.Value:
 type subprocessDepthKey struct{}
 func withSubprocessDepth(ctx, depth) context.Context
 func subprocessDepth(ctx) int

 Subprocess executor (implements PluginExecutor):
 type subprocessExecutor struct { orch *Orchestrator }
 func (s *subprocessExecutor) Execute(ctx, call) ToolResult

 Mini agent loop — runSubprocess(ctx, req, depth):
 - Creates ephemeral message list (no session persistence)
 - Builds scoped system prompt with only allowed tools (excluding _subprocess to prevent fork bombs)
[...]

PR #106, opened 2026-04-27
feat(orchestrator): add subprocess (sub-agent) forking support (#105)

OpakAlex:
Add Unix-inspired subprocess system where the main LLM agent loop can fork focused child processes for sub-tasks (vector search, research, multi-step tool usage). Children run their own mini agent loop with ephemeral context and scoped tools, returning results to the parent.

### DIFF
--- a/internal/orchestrator/subprocess.go
+++ b/internal/orchestrator/subprocess.go
@@ -1,0 +1,294 @@
+package orchestrator
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/opentalon/opentalon/internal/provider"
+)
+
+// SubprocessConfig configures the subprocess (sub-agent) system.
+type SubprocessConfig struct {
+	Enabled        bool          // master switch
+	MaxDepth       int           // max nesting depth (default 2, hard cap 3)
+	MaxIterations  int           // default iterations per child (default 5, hard cap 10)
+	DefaultTimeout time.Duration // per-subprocess timeout (default 60s)
+}
+
+// subprocessRequest is parsed from the LLM's _subprocess.run tool call args.
+type subprocessRequest struct {
+	Task          string
+	AllowedTools  []string
+	MaxIterations int
+}
+
+// subprocessDepthKey is the context key for tracking subprocess nesting depth.
+type subprocessDepthKey struct{}
+
+func withSubprocessDepth(ctx context.Context, depth int) context.Context {
+	return context.WithValue(ctx, subprocessDepthKey{}, depth)
+}
+
+func subprocessDepth(ctx context.Context) int {
+	v, _ := ctx.Value(subprocessDepthKey{}).(int)
+	return v
+}
+
+// subprocessExecutor implements PluginExecutor for the built-in _subprocess plugin.
+type subprocessExecutor struct {
+	orch *Orchestrator
+}
+
+func (s *subprocessExecutor) Execute(ctx context.Context, call ToolCall) ToolResult {
+	req, err := parseSubprocessRequest(call.Args)
+	if err != nil {
+		return ToolResult{CallID: call.ID, Error: err.Error()}
+	}
+
+	currentDepth := subprocessDepth(ctx)
+	maxDepth := s.orch.subprocessConfig.MaxDepth
+	if maxDepth <= 0 {
+		maxDepth = 2
+	}
+	if maxDepth > 3 {
+		maxDepth = 3
+	}
+	if currentDepth >= maxDepth {
+		return ToolResult{
+			CallID: call.ID,
+			Error:  fmt.Sprintf("subprocess depth limit reached (%d)", maxDepth),
+		}
+	}
+
+	result, err := s.orch.runSubprocess(ctx, req, currentDepth+1)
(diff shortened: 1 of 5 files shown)


zhisme (commented):
Overall solid. Depth limiting, fork-bomb prevention, allowlist filtering, and timeout all correct. A few issues inline.

internal/orchestrator/subprocess_test.go
```diff
@@ -0,0 +1,410 @@
...
+
+	orch := setupSubprocessOrchestrator(llm)
+	orch.subprocessConfig.MaxIterations = 3
+
+	req := subprocessRequest{Task: "keep searching", MaxIterations: 3}
+	result, err := orch.runSubprocess(context.Background(), req, 1)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if !strings.Contains(result.Response, "iteration limit") {
```
zhisme, line 138: Dead variable — `callCount` is assigned but never used. Remove it and `_ = callCount`.

internal/orchestrator/subprocess_test.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+}
+
+func TestSubprocessBlocksRecursion(t *testing.T) {
+	// Subprocess tries to call _subprocess.run — should be blocked.
+	llm := &fakeLLM{responses: []string{
+		`[tool_call]
+{"tool": "_subprocess.run", "args": {"task": "recursive task"}}
+[/tool_call]`,
+		"I can't spawn subprocesses from here.",
```
zhisme, line 178: Dead code — `runSubprocess` result discarded with `_ = result; _ = err`. The comment even admits the depth check is in the executor, not here. Delete lines 188–193.

internal/orchestrator/subprocess_test.go
```diff
@@ -0,0 +1,410 @@
...
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if tt.check != nil {
+				tt.check(t, req)
+			}
+		})
```
zhisme, line 323: Weak assertion — only checks `result.Response != ""`. Does not verify the subprocess actually received an error for the `_subprocess` call. Would pass silently if blocking were broken. Assert the LLM saw an error, or check `result.Iterations`.

internal/orchestrator/subprocess.go
```diff
@@ -0,0 +1,294 @@
...
+	if maxDepth <= 0 {
+		maxDepth = 2
+	}
+	if maxDepth > 3 {
+		maxDepth = 3
+	}
+	if currentDepth >= maxDepth {
+		return ToolResult{
+			CallID: call.ID,
+			Error:  fmt.Sprintf("subprocess depth limit reached (%d)", maxDepth),
```
zhisme: Error message is ambiguous — `%d` is the limit but reads like current depth. Suggest: `"subprocess depth limit reached (depth %d of %d)", currentDepth+1, maxDepth`.

internal/orchestrator/subprocess.go
```diff
@@ -0,0 +1,294 @@
...
+
+			toolResult := o.executeCall(ctx, call)
+
+			if o.pluginCallObserver != nil {
+				o.pluginCallObserver.ObservePluginCall(call.Plugin, call.Action, toolResult.Error != "", 0, 0)
+			}
+
+			messages = append(messages,
+				provider.Message{Role: provider.RoleAssistant, Content: formatToolCallMessage(call)},
+				provider.Message{Role: provider.RoleUser, Content: o.guard.WrapContent(toolResult)},
```
zhisme, line 173: `ObservePluginCall` called with `0, 0` for input/output token counts. Subprocess has `resp.Usage` available — pass actual counts or this skews per-plugin metrics aggregated upstream.

internal/orchestrator/subprocess.go
```diff
@@ -0,0 +1,294 @@
...
+		allowSet[t] = true
+	}
+	hasAllowlist := len(allowSet) > 0
+
+	// Don't list preparer/guard actions.
+	preparerAction := make(map[string]bool)
+	for _, prep := range o.preparers {
+		preparerAction[prep.Plugin+"."+prep.Action] = true
+	}
+	for _, g := range o.guards {
```
zhisme, line 205: Misleading name — this map holds both preparer and guard actions but is called `preparerAction`. Rename to `internalActions` or `skipActions`.

internal/orchestrator/subprocess.go
```diff
@@ -0,0 +1,294 @@
...
+
+		fmt.Fprintf(&sb, "## %s\n%s\n", cap.Name, cap.Description)
+		for _, action := range visibleActions {
+			fmt.Fprintf(&sb, "- %s.%s: %s\n", cap.Name, action.Name, action.Description)
+			for _, p := range action.Parameters {
+				req := ""
+				if p.Required {
+					req = " (required)"
+				}
+				fmt.Fprintf(&sb, "  - %s: %s%s\n", p.Name, p.Description, req)
```
zhisme: `req := ""` shadows the outer function parameter `req subprocessRequest`. Rename to `requiredSuffix` or `reqMark`.

PR #126, opened 2026-04-28
feat: show status labels in Slack during processing

OpakAlex:
Add transient status indicators ("Thinking...", "Calling plugin.action...") that appear in the Slack message while the bot is working. The status is displayed as italic text and gets replaced by the first real content chunk or the next status update.

- Add SendStatus method to StreamWriter
- Send "Thinking..." before first LLM call
- Send "Calling plugin.action..." before each tool execution

### DIFF
--- a/pkg/channel/stream.go
+++ b/pkg/channel/stream.go
@@ -185,6 +185,43 @@ func (sw *StreamWriter) flush(ctx context.Context) {
 	}
 }
 
+// SendStatus shows a transient status label (e.g. "Thinking...", "Calling timly.list-items...")
+// on the channel. The status is displayed as a standalone message that will be replaced
+// by the first real content chunk or the next status update. Only works on updatable channels.
+func (sw *StreamWriter) SendStatus(ctx context.Context, label string) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	msg := OutboundMessage{
+		ConversationID: sw.convID,
+		ThreadID:       sw.threadID,
+		Content:        "_" + label + "_",
+		Metadata:       sw.cloneMetadata(),
+	}
+
+	if sw.flushed && sw.messageID != "" {
+		if uch, ok := sw.ch.(UpdatableChannel); ok {
+			if err := uch.SendUpdate(ctx, sw.messageID, msg); err != nil {
+				slog.Debug("status update failed", "error", err)
+			}
+			return
+		}
+	}
+
+	if !sw.flushed {
+		if uch, ok := sw.ch.(UpdatableChannel); ok {
+			msgID, err := uch.SendAndCapture(ctx, msg)
+			if err != nil {
+				slog.Debug("status send failed", "error", err)
+				return
+			}
+			sw.messageID = msgID
+			sw.flushed = true
+			return
+		}
+	}
+}
+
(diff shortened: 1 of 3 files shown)


zhisme (commented):
## Review

**Critical bug (point 3 below).** Lock-over-I/O and missing test for the failure case also need attention.

pkg/channel/stream.go
```diff
@@ -185,6 +185,43 @@ func (sw *StreamWriter) flush(ctx context.Context) {
...
+// by the first real content chunk or the next status update. Only works on updatable channels.
+func (sw *StreamWriter) SendStatus(ctx context.Context, label string) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	msg := OutboundMessage{
+		ConversationID: sw.convID,
+		ThreadID:       sw.threadID,
+		Content:        "_" + label + "_",
+		Metadata:       sw.cloneMetadata(),
```
zhisme, line 199: Lock held across `SendAndCapture` and `SendUpdate` (network I/O). Blocks concurrent `OnChunk` calls. Copy needed state, release lock, then do I/O.

pkg/channel/stream.go
```diff
@@ -185,6 +185,43 @@ func (sw *StreamWriter) flush(ctx context.Context) {
...
+				slog.Debug("status update failed", "error", err)
+			}
+			return
+		}
+	}
+
+	if !sw.flushed {
+		if uch, ok := sw.ch.(UpdatableChannel); ok {
+			msgID, err := uch.SendAndCapture(ctx, msg)
+			if err != nil {
```
zhisme, line 214: **Bug:** after real content is flushed, `sw.flushed == true` and `sw.messageID` holds the real message. A subsequent tool-call status hits this branch and overwrites real content with "Calling…".

Need to distinguish status-only messages from content messages — e.g. a `statusMessageID` field separate from `messageID`, or a `contentFlushed bool` flag. When `contentFlushed`, skip the update entirely (status after real content is pointless).

pkg/channel/stream.go
```diff
@@ -185,6 +185,43 @@ func (sw *StreamWriter) flush(ctx context.Context) {
...
+				slog.Debug("status send failed", "error", err)
+				return
+			}
+			sw.messageID = msgID
+			sw.flushed = true
+			return
+		}
+	}
+}
+
```
zhisme, line 224: Non-updatable channel silently no-ops. Add a `slog.Debug` here — helps diagnose why statuses don't appear on certain channel types.

pkg/channel/stream_test.go
```diff
@@ -152,6 +152,41 @@ func TestStreamWriterWithUpdatableChannel(t *testing.T) {
...
+	}
+
+	// Real content replaces the status.
+	sw.OnChunk(ctx, "The answer is 42.", true)
+	lastUpdate := ch.updates[len(ch.updates)-1]
+	if lastUpdate.Content != "The answer is 42." {
+		t.Errorf("final content = %q, want real answer", lastUpdate.Content)
+	}
+}
+
```
zhisme, line 189: Missing test: status sent *after* real content already flushed. That's the scenario where the bug in `SendStatus` manifests — `OnChunk` first, then `SendStatus`, verify real content is NOT overwritten.

PR #149, opened 2026-04-29
feat/setup integration tests

### LINKED ISSUE #118: Integration test with cheapest LLM interaction
Design test that can trigger opentalon against real (cheapest) LLM and verify that new changes does not break build in production

- create VCR for replaying against recent code changes
- structural test with real api calls
- eval with real api keys before releasing to track regression against newer models

zhisme:
closes #118


zhisme (commented):
## Integration Test Framework Review

**What it does:** Extracts shared scenarios into YAML, adds a `scenarios_hash` field to VCR cassettes, adds structural integration tests (real API, temperature=0), adds an eval framework with baseline tracking, and wires up two CI workflows.

---

### Issues

**1. `vcr-release.yml` pushes directly to master**

```yaml
git push origin master
```

Bot commits non-deterministic content (new cassette recordings) straight to the default branch on every release, bypassing any PR/review gate. If re-recording produces unexpected changes, they land silently. Better: open a PR or push to a branch and require review.

**2. Eval baselines never exist on CI fresh checkouts → regression guard is a no-op**

`internal/eval/eval_test.go:467-485` — first run saves baseline, second run checks it. On a release CI run with a fresh checkout, the baseline file for that tag doesn't exist, so it saves and exits. The regression check only fires if the *same tag* was already run and saved. Baselines need to be committed to the repo for this to work.

**3. `passThreshold = 1.0` in `eval_test.go:313`**

100% threshold means any scenario flake fails CI. Even with temperature=0, model outputs aren't bit-for-bit stable across versions. This will produce noisy release failures. Consider 0.9 or make it configurable.

**4. `zeroTempLLM` duplicated**

`internal/eval/eval_test.go:321` and `internal/orchestrator/integration_test.go:572` — identical struct/method. Move to a shared `internal/testutil` or the `scenarios` package.

**5. `LoadScenarios` reads disk, `Hash()` reads embedded FS**

`scenarios.go:25` uses `//go:embed` (compile-time), `LoadScenarios` at line 83 uses `os.ReadDir` (runtime). Adding a YAML and running tests without recompiling will load the new scenario but hash the old content. Not a CI issue, but confusing locally.

**6. Cassette name collision risk**

`CassetteName` (`scenarios.go:88`) does `strings.ToLower + space→underscore`. "Direct Response" and "direct response" map to the same file — silent overwrite. Add a uniqueness check in `LoadScenarios`.

---

### Minor

- `vcrRunResult` (`vcr_test.go`) and `orchRunResult` (`integration_test.go`) are near-identical converters.
- Nested subtests in `TestEval` / `TestIntegrationScenarios` don't call `t.Parallel()` — sequential API calls on slow paths.

---

### What's good

- YAML-driven scenarios cleanly decouple test data from code.
- `scenarios_hash` staleness check in VCR is the right call.
- `assertStructural` catching parser leakage (`[tool_call]` in response) is good defensive testing.
- `TestVCRScenarios` skip-on-missing-cassette pattern is clean.

PR #151, opened 2026-04-29
Feat/session ownership

OpakAlex:


zhisme (commented):
Messages normalization is solid work. Four issues worth fixing before merge: race in AddMessage, non-transactional SetSummary, silent backfill failures, and DSN leaking to all plugins. Details inline.

internal/state/store/session.go
```diff
@@ -82,93 +84,83 @@ func (s *SessionStore) Create(id string) *state.Session {
...
-		string(newMessagesJSON), updatedAt, id)
+	now := time.Now().UTC().Format(time.RFC3339)
+
+	// Get next seq number for this session.
+	var maxSeq int
+	_ = s.db.SQLDB().QueryRowContext(ctx,
+		d.Rebind(`SELECT COALESCE(MAX(seq), 0) FROM messages WHERE session_id = ?`), id,
+	).Scan(&maxSeq)
+	seq := maxSeq + 1
+
```
zhisme: Race condition: SELECT MAX(seq) and INSERT run in separate statements with no transaction. Two concurrent AddMessage calls read the same maxSeq and produce a seq collision against the UNIQUE INDEX — one INSERT fails and the message is lost. The old code used BeginExclusive to prevent this.

Fix: wrap the SELECT + INSERT in an exclusive transaction, or use a single atomic statement:
```sql
INSERT INTO messages (session_id, seq, role, content, created_at)
SELECT ?, COALESCE(MAX(seq), 0) + 1, ?, ?, ?
FROM messages WHERE session_id = ?
```
(SQLite supports this; Postgres needs a CTE or advisory lock.)

internal/state/store/session.go
```diff
@@ -55,16 +58,15 @@ func (s *SessionStore) Get(id string) (*state.Session, error) {
...
 
 // Create inserts a new session. If id already exists (e.g. race), returns existing session from DB.
-func (s *SessionStore) Create(id string) *state.Session {
+func (s *SessionStore) Create(id, entityID, groupID string) *state.Session {
 	now := time.Now().UTC().Format(time.RFC3339)
 	_, err := s.db.SQLDB().Exec(
-		s.db.Dialect().Rebind(`INSERT INTO sessions (id, messages, summary, active_model, metadata, created_at, updated_at) VALUES (?, ?, ?, ?, ?, ?, ?)`),
-		id, "[]", "", "", "{}", now, now)
+		s.db.Dialect().Rebind(`INSERT INTO sessions (id, messages, summary, active_model, metadata, entity_id, group_id, created_at, updated_at) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)`),
+		id, "[]", "", "", "{}", entityID, groupID, now, now)
```
zhisme: `Create` still writes `messages = "[]"` into the old `sessions.messages` column, but `Get` now reads from the `messages` table. The column is vestigial after migration 005. Either stop writing it here, or add a migration to drop it — otherwise any direct `SELECT messages FROM sessions` query returns stale data and this will confuse future ops/migrations.

internal/state/store/session.go
```diff
@@ -82,93 +84,83 @@ func (s *SessionStore) Create(id string) *state.Session {
...
+	d := s.db.Dialect()
+	now := time.Now().UTC().Format(time.RFC3339)
 
-func (s *SessionStore) persist(sess *state.Session) error {
-	messagesJSON, err := json.Marshal(sess.Messages)
-	if err != nil {
-		return fmt.Errorf("session persist: marshal messages: %w", err)
+	// Delete all existing messages for this session.
+	if _, err := s.db.SQLDB().ExecContext(ctx,
+		d.Rebind(`DELETE FROM messages WHERE session_id = ?`), id); err != nil {
```
zhisme, line 143: `SetSummary` is not transactional: DELETE then N INSERTs run as separate statements. A crash or error mid-insert leaves the session with a committed summary but partial message history. The orchestrator summarization path relies on this being atomic.

Wrap in a transaction:
```go
tx, err := s.db.SQLDB().BeginTx(ctx, nil)
// DELETE, INSERTs, UPDATE all inside tx
tx.Commit()
```

internal/state/store/db.go
```diff
@@ -206,3 +211,37 @@ func migrationSQL(name string) (string, error) {
...
 	return string(data), nil
 }
+
+// backfillSessionOwnership populates entity_id and group_id on sessions
+// created before migration 004 by parsing the session ID (entityID:channel:conv...).
+func (d *DB) backfillSessionOwnership() {
+	rows, err := d.db.Query(d.dialect.Rebind(`SELECT id FROM sessions WHERE entity_id = ''`))
+	if err != nil {
+		return
+	}
```
zhisme, line 223: Backfill failure is completely silent. If the initial query fails, we return with no log — the caller and operator have no idea backfill was skipped. At minimum:
```go
if err != nil {
    slog.Warn("session ownership backfill failed", "error", err)
    return
}
```
Same for the per-row update errors on line 52.

cmd/opentalon/main.go
```diff
@@ -193,8 +193,25 @@ func main() {
...
+		// Auto-inject DB connection info so plugins can query the state store.
+		if _, ok := pluginCfg["__db_driver"]; !ok {
+			driver := cfg.State.DB.Driver
+			if driver == "" {
+				driver = "sqlite"
+			}
+			pluginCfg["__db_driver"] = driver
+			if driver == "postgres" {
+				pluginCfg["__db_dsn"] = cfg.State.DB.DSN
+			} else if dataDir != "" {
```
zhisme: The full Postgres DSN (including password) is now auto-injected into every enabled plugin's config as `__db_dsn`. Any plugin — including third-party ones — gets raw DB credentials and can query the state store directly, bypassing all orchestrator-level access control.

Is this intentional? If so, it needs to be an explicit opt-in (e.g. a per-plugin `allow_db_access: true` flag), not a silent global inject. The comment treats this as obviously safe but it is not.

Issue #26, opened 2026-03-06
feat: add `make clean-cache` to clear cached plugins, skills, and channels


zhisme:
## Problem

When a skill, plugin, or channel is updated upstream (e.g. a bug fix pushed to `master`), OpenTalon keeps using the cached/locked version from `state.data_dir`. Users must manually delete cached directories and lock file entries to force a re-download. This is tedious and error-prone.

## Proposal

Add a `make clean-cache` (or `./opentalon -clean all`) command that:

1. Removes all cached skills, plugins, and channels under `state.data_dir` (default `~/.opentalon`)
2. Clears `skills.lock`, `plugins.lock`, `channels.lock`, and `lua_plugins.lock`
3. Next run re-downloads and rebuilds everything from the configured refs

Selective cleaning:
- `make clean-skills` — only skills
- `make clean-plugins` — only plugins  
- `make clean-channels` — only channels
- `make clean-lua-plugins` — only Lua plugins
- `make clean-cache` — everything

## Context

Discovered while testing [opentalon/brave-plugin](https://github.com/opentalon/brave-plugin) — after pushing a fix, the cached version persisted and required manual cleanup.

PR #198, opened 2026-05-08
feat: pipeline step-output substitution + typed-args end-to-end

mscdit:
## TL;DR

Multi-step planner pipelines could not reference values produced by earlier steps. The planner LLM emitted placeholders like `{{step1.output.containers[0].id}}` based on common templating conventions, but the executor had no substitution layer — the literal string went to the wire, the downstream MCP server rejected with "property '#/item_id' of type string did not match the following type: integer", and the agent loop had no path to recover.

This PR closes the gap end-to-end and cleans up the four collocated bugs that fall out of the same diagnosis.

## Reproduction (live trace, current master)

User asked: *"Can you reserve the Tesla in the Berlin garage from 1 June to 5 June?"*

```
2026/05/08 19:47:33 mcp-plugin: server timly: CallTool begin tool="schedule-item" arg_count=4
  args=[
    container_id={{step1.output.containers[0].id}}(string),
    end_date=2024-06-05T23:59:59Z(string),
    item_id={{step2.output.items[0].id}}(string),
    start_date=2024-06-01T00:00:00Z(string)
  ]
2026/05/08 19:47:33 mcp-plugin: server timly: CallTool tool="schedule-item" rpc error:
  Invalid params: Invalid arguments:
  The property '#/item_id'      of type string did not match the following type: integer
  The property '#/container_id' of type string did not match the following type: integer
```

The placeholder strings reach the MCP server unsubstituted. (Note: this trace is only readable thanks to opentalon/mcp-plugin#26 surfacing JSON-RPC `error.data` — without that, "Invalid params" had no detail.)

## What's wrong, in one place

| File | Bug |
|---|---|
| `internal/pipeline/executor.go:Run` | `step.Command.Args` goes to runner unchanged — no substitution. `step.Result.Data["output"] = runResult.Content` — text only, no JSON path. |
| `internal/orchestrator/orchestrator.go:executePipeline` | runner forwards `Content`, drops `StructuredContent` — no traversable shape. |
| `internal/pipeline/planner.go:parsePlanResponse` | `args[k] = fmt.Sprintf("%v", v)` — `float64(2_037_838)` → `"2.037838e+06"`. |
| `internal/prompts/planner_preamble.txt` | no contract for chaining syntax — LLM invents one. |

### 1. Step-output substitution (`internal/pipeline/substitution.go`)

New engine resolving `{{<step-id>.output.<json-path>}}` against the pipeline context. JSON-path walk: dot-separated keys + `[N]` for array indices.

When a string IS one placeholder, the resolved value's type is preserved (an integer id stays integer); when embedded into surrounding text, the value is stringified with no scientific-notation surprise. Errors include the offending arg key, the failed segment, and the available alternatives — actionable diagnostics that the agent loop can surface verbatim.

### 2. Typed args throughout the pipeline

`PluginCommand.Args` is now `map[string]any` (was `map[string]string`). The planner stops doing `fmt.Sprintf("%v", v)` — JSON numbers stay `float64`, booleans stay `bool`, nested values keep shape. The collapse to `map[string]string` happens once, at the orchestrator/runner boundary, where typed-aware rendering keeps integer-valued floats in plain decimal form.

### 3. Structured output into the pipeline context

`StepRunResult.StructuredContent` is added; the orchestrator's pipeline runner forwards it; the executor parses it as JSON and stores the parsed value at `<step>.output`. References traverse real data, not text. Tools without a structured channel fall back to text — references against text fail with "expected object, got string", which is the right error to surface.

### 4. Wire-boundary stringification (`internal/orchestrator/pipeline_args.go`)

`pipelineArgsToWire` collapses typed args to `map[string]string` for ToolCall and the gRPC plugin protocol. Integer-valued floats render as plain decimals (regression-tested up to 10^18), arrays and objects round-trip via JSON so the downstream plugin can re-coerce per its declared input schema.

### 5. Planner prompt contract

`planner_preamble.txt` now documents the substitution syntax with a concrete example (list-containers + list-items + schedule-item). The example response keys (`containers[]`, `items[]`) match what real tools return, with an explicit reminder to check the action description before guessing path names.

## Why one PR, not five

Each piece is a hard prerequisite for the next:

```
substitution needs typed values
              ⇕
typed values need a typed-aware wire boundary (else %v breaks IDs)
              ⇕
substitution needs structured output (else paths can't traverse)
              ⇕
the LLM must know the syntax (else it invents one)
```

Splitting them ships a partial fix that demonstrably worsens behaviour in the agent loop (the LLM saw the rich error surfaced by opentalon/mcp-plugin#26 and tried to self-correct, but had no syntax to use; or it had syntax but no real data behind it). They land together or not at all.

## Failure modes (all fail-fast, all actionable)

- Unresolved step: `step "step9" produced no output`
- Key not found: `key "items" not found (available: containers, pagination)`
- Out-of-range index: `index [5] out of range (length 1)`
- Type mismatch: `expected object, got number`
- Malformed path: `unclosed '[' in path "items[0"`
- All errors prefixed with the offending arg key

Substitution failures put the step straight to `StepFailed` without a runner attempt; `FailFast` halts the pipeline. No retries on deterministic plan errors.

## Tests (~430 LOC of new test code)

```
internal/pipeline/substitution_test.go
  TestSplitPath:                              12 cases (paths + diagnostics)
  TestResolvePathTraversal:                    4 cases
  TestResolvePathDiagnostics:                  4 cases (each error class)
  TestResolveArgsSoloPlaceholderPreservesType
  TestResolveArgsInterpolation
  TestResolveArgsLargeIntegerNoScientificNotation  ← regression for original bug
  TestResolveArgsNoPlaceholders                                ← backward-compat
  TestResolveArgsNestedValues                                  ← recurse []/map
  TestResolveArgsErrorPaths:                   4 cases (each error surfaces arg key)
  TestParseStepOutput*:                        3 cases (structured / fall-back / malformed)

internal/pipeline/executor_test.go (additions)
  TestExecutorSubstitutesStepReferences        ← Tesla-reservation scenario, full E2E
  TestExecutorRecordsSubstitutedArgsInTrace
  TestExecutorFailsFastOnUnresolvedReference   ← NO retry on plan error
  TestExecutorContextStoresStructuredOutput

internal/orchestrator/pipeline_args_test.go
  TestArgToWireString:                        17 cases incl. NaN / ±Inf / json.Number
  TestPipelineArgsToWireMixed                  ← realistic post-substitution shape

internal/pipeline/planner_test.go (addition)
  TestBuildPlannerPrompt_DocumentsChainingSyntax  ← pin the contract in the prompt
```

`go test ./...` (excl. VCR — see below): all green. `golangci-lint run ./internal/pipeline/... ./internal/orchestrator/...`: 0 issues.

## ⚠️ VCR cassettes need re-recording

`internal/prompts/planner_preamble.txt` change rotates `prompts.Hash()`, which is the staleness guard for VCR cassettes. The following fail with the documented "stale cassette" error and require `make vcr-record-all` against the live LLM (`ANTHROPIC_API_KEY` + `OPENROUTER_API_KEY`):

- `TestVCRScenarios/{direct_response,single_tool_call,...}` (both providers)
- `TestVCRSelfIntroduction`
- `TestVCRPipelineConfirmation`
- `TestVCRFormatSlack`
- `TestVCRFormatTelegram`

This is the documented behaviour — the hash guard exists to force a manual re-record on prompt changes — and matches the precedent set in #174 (planner-prompt change, cassette hashes rebumped). Maintainer with API keys re-records before merge.

## Backward-compatibility

- Steps without placeholders pass through unchanged.
- Tools without `StructuredContent` fall back to the existing text-only context.
- The wire format (`proto/plugin.proto`) is unchanged — args still reach the plugin as `map<string,string>`.
- The planner prompt addition is additive.

## Companion / context

- opentalon/mcp-plugin#26 (already merged, tagged in opentalon v0.0.14) — surfaced the JSON-RPC `error.data` field that made this bug visible. Together they restore the "tool error → LLM self-correct" loop end-to-end.

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -870,7 +870,7 @@ func (o *Orchestrator) Run(ctx context.Context, sessionID, userMessage string, f
 					ID:     fmt.Sprintf("planner-%s-%s", step.Command.Plugin, step.Command.Action),
 					Plugin: step.Command.Plugin,
 					Action: step.Command.Action,
-					Args:   step.Command.Args,
+					Args:   pipelineArgsToWire(step.Command.Args),
 				}
 				toolResult := o.executeCall(ctx, call)
 				if toolResult.Error == "" {
@@ -1756,13 +1756,17 @@ func (o *Orchestrator) runGuardPlugins(ctx context.Context, messages []provider.
 
 func (o *Orchestrator) executePipeline(ctx context.Context, sessionID string, p *pipeline.Pipeline) (*RunResult, error) {
 	log := logger.FromContext(ctx)
-	runner := func(ctx context.Context, pluginName, action string, args map[string]string) pipeline.StepRunResult {
-		log.Debug("pipeline executing step", "plugin", pluginName, "action", action, "args", args)
+	runner := func(ctx context.Context, pluginName, action string, args map[string]any) pipeline.StepRunResult {
+		// Stringify typed args at the wire boundary; ToolCall.Args is
+		// map[string]string by gRPC contract. The downstream plugin re-
+		// coerces per its declared input schema.
+		wireArgs := pipelineArgsToWire(args)
+		log.Debug("pipeline executing step", "plugin", pluginName, "action", action, "args", wireArgs)
 		call := ToolCall{
 			ID:     fmt.Sprintf("pipeline-%s-%s", pluginName, action),
 			Plugin: pluginName,
 			Action: action,
-			Args:   args,
+			Args:   wireArgs,
 		}
 		result := o.executeCall(ctx, call)
 		if result.Error != "" {
@@ -1774,7 +1778,11 @@ func (o *Orchestrator) executePipeline(ctx context.Context, sessionID string, p 
 			}
 			log.Debug("pipeline step succeeded", "plugin", pluginName, "action", action, "result", preview)
 		}
-		return pipeline.StepRunResult{Content: result.Content, Error: result.Error}
+		return pipeline.StepRunResult{
+			Content:           result.Content,
+			StructuredContent: result.StructuredContent,
+			Error:             result.Error,
+		}
 	}
 	executor := pipeline.NewExecutor(runner, p.Config)
 	execResult, err := executor.Run(ctx, p)
@@ -1784,7 +1792,8 @@ func (o *Orchestrator) executePipeline(ctx context.Context, sessionID string, p 
 
 	log.Debug("pipeline execution done", "success", execResult.Success, "steps_executed", len(execResult.Steps))
 
-	// Record step results in session history
+	// Record step results in session history. Args are stringified at the
+	// recording boundary so the trace mirrors what hit the wire.
 	var toolCalls []ToolCall
 	var toolResults []ToolResult
 	for _, es := range execResult.Steps {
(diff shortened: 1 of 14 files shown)


OpakAlex:
https://github.com/opentalon/opentalon/pull/new/feat/pipeline-templated-step-refs

PR #224, opened 2026-05-11
feat: debounce rapid messages into single LLM call

OpakAlex:
## Summary
- Adds opt-in message debouncing: when the user sends multiple messages rapidly, they're merged into a single `Run()` call instead of burning an LLM round per message
- Disabled by default (0ms window) — enable via `registry.SetDebounceWindow(800 * time.Millisecond)`
- Confirmation signals (`metadata["confirmation"]`) and typing indicators bypass debounce for instant processing

## How it works
- Messages for the same session are buffered for a configurable window (e.g. 800ms)
- Each new message resets the timer
- When the timer fires, all buffered messages are merged: content joined with `\n`, metadata merged (last wins), files concatenated
- Different sessions debounce independently

## Test plan
- [x] `go test ./internal/channel/ -count=1` passes (6 new tests)
- [x] All existing tests pass (debounce disabled by default)
- [x] `golangci-lint run` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,6 +618,25 @@ Session A msg2 ···············► (waits for msg1 to finish) ─�
 
 Set `max_concurrent_sessions` to match your LLM provider's rate limits and expected load. A value of `1` (default) gives the original sequential behaviour with no behaviour change needed for existing deployments.
 
+## Message Debouncing
+
+When a user sends multiple messages rapidly (e.g. "yes", "also add barcode 123", "set price to 50"), each message normally triggers a separate LLM call. With large context windows (~98k tokens), this wastes tokens and produces disjointed responses.
+
+Enable debouncing to merge rapid messages into a single LLM call:
+
+```yaml
+orchestrator:
+  debounce_window: "800ms"   # collect messages for 800ms, then send as one (default: "0" = disabled)
+```
+
+**How it works:**
+
+- When a message arrives, the debouncer waits for the configured window (e.g. 800ms)
+- If more messages arrive within the window, the timer resets
+- When the timer fires, all buffered messages are merged (content joined with newlines) and processed as one request
+- Different sessions debounce independently
+- **Confirmation signals** (`approve`/`reject`) bypass debounce and execute immediately
+
 ### Debugging
 
 Set `LOG_LEVEL=debug` to see pipeline decisions, planner LLM calls, step execution, and retry attempts — all prefixed with `[pipeline]`.
--- a/README.md
+++ b/README.md
@@ -635,7 +635,7 @@ orchestrator:
 - If more messages arrive within the window, the timer resets
 - When the timer fires, all buffered messages are merged (content joined with newlines) and processed as one request
 - Different sessions debounce independently
-- **Confirmation signals** (`approve`/`reject`) bypass debounce and execute immediately
+- **Confirmation signals** (any message with non-empty `confirmation` metadata) and typing indicators bypass debounce and execute immediately
 
 ### Debugging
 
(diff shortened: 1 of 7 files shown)


zhisme (commented):
Reviewed debounce implementation. Issues found, ordered by severity.

internal/channel/debounce.go
```diff
@@ -0,0 +1,134 @@
...
+	if buf.timer != nil {
+		buf.timer.Stop()
+	}
+	buf.timer = time.AfterFunc(d.window, func() {
+		d.flush(sessionKey)
+	})
+
+	return true
+}
+
```
zhisme, line 91: Timer leak on shutdown. Pending timers fire after `r.ctx` is cancelled — buffered goroutines call `processMessage` on a done context. Need a `stop()` on `sessionDebouncer` that cancels all pending timers; call it when the dispatch loop exits.

internal/channel/registry.go
```diff
@@ -38,13 +39,23 @@ type Registry struct {
...
+		ctx:            ctx,
+		cancel:         cancel,
 	}
 }
 
+// SetDebounceWindow overrides the default debounce window for message batching.
+// Set to 0 to disable debouncing (each message dispatched immediately).
+// Must be called before any channels are registered.
+func (r *Registry) SetDebounceWindow(d time.Duration) {
+	r.mu.Lock()
```
zhisme, line 54: "Must be called before any channels are registered" is unenforced. `debouncer` is created once per `dispatch()` goroutine at channel-register time. Calling `SetDebounceWindow` after that silently has no effect. Either panic/error if called late, or create the debouncer lazily inside `submit` from `r.debounceWindow`.

internal/channel/registry.go
```diff
@@ -159,93 +190,86 @@ func (r *Registry) dispatch(ch pkg.Channel, inbox <-chan pkg.InboundMessage) {
 			dedup, dedupTTL := r.dedup, r.dedupTTL
```
zhisme, line 202: Dedup key uses `msg.Timestamp` from original message, before merge. After debounce, merged message carries `last.Timestamp`. So a pod-duplicate of the last original message could be incorrectly deduplicated post-merge if it arrives after the timer fires. The ordering (dedup before debounce) is correct for pod-level suppression — add a comment explaining it, because the ordering is load-bearing.

internal/channel/debounce_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Now it should be dispatched.
+	time.Sleep(50 * time.Millisecond)
+	mu.Lock()
+	defer mu.Unlock()
+	if len(dispatched) != 1 {
+		t.Fatalf("expected 1 dispatch, got %d", len(dispatched))
+	}
+	if dispatched[0].Content != "solo" {
+		t.Errorf("content = %q, want %q", dispatched[0].Content, "solo")
```
zhisme, line 117: Flaky timing. 20ms assert-not-dispatched window is too tight on loaded CI. Use `3 * window` for the final sleep, and either use a signalling channel or accept that the early-assert is inherently racy. Suggest:\n```go\ntime.Sleep(3 * window)\n// assert dispatched\n```\nDrop the intermediate not-yet-dispatched check entirely — it's not worth the flakiness.

README.md
```diff
@@ -618,6 +618,25 @@ Session A msg2 ···············► (waits for msg1 to finish) ─
...
+
+Enable debouncing to merge rapid messages into a single LLM call:
+
+```yaml
+orchestrator:
+  debounce_window: "800ms"   # collect messages for 800ms, then send as one (default: "0" = disabled)
+```
+
+**How it works:**
+
```
zhisme, line 633: README says `approve`/`reject` bypass debounce, but code checks `msg.Metadata["confirmation"] != ""` — any non-empty value bypasses. Docs should match: "any non-empty `confirmation` metadata value bypasses debounce".

